PR #131, opened 2018-08-06
adds some tests for the universe contracts and fixes one bug in solid…

huesforalice:
…ity code

### DIFF
--- a/contracts/Universe.sol
+++ b/contracts/Universe.sol
@@ -175,6 +175,7 @@ contract Universe is
         uint256 idx;
         while(idx < interfaceIds.length) {
             setCollectionPrivate(interfaceIds[idx], instance, set);
+            idx += 1;
         }
     }
 
--- a/test/Universe.js
+++ b/test/Universe.js
@@ -154,16 +154,16 @@ contract(
     });
 
     it("should set singleton without convenience method", async () => {
-      const some_id = "0x63452ff6";
+      const someId = "0x63452ff6";
       // set neumark to address other
-      let tx = await universe.setSingleton(some_id, other, {
+      const tx = await universe.setSingleton(someId, other, {
         from: universeManager,
       });
-      expectSingletonSetEvent(tx, some_id, other, ZERO_ADDRESS);
-      expect(await universe.getSingleton(some_id)).to.eq(other);
-      expect(await universe.getManySingletons([some_id])).to.deep.eq([other]);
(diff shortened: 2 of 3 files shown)


huesforalice:
@rudolfix I have two questions:
"should set singleton without convenience method" -> I'm not sure what to test here exactly
"should remove interface from collection" -> This, I think, is now tested in the tests above, please let me know if you see this differently.

rudolfix:
Hey did you upgrade linter? solium? The advice to always provide message for require is very bad. it costs a lot and contracts grow very quickly.

also javascript linter fails

contracts/Universe.sol
```diff
@@ -175,6 +175,7 @@ contract Universe is
         uint256 idx;
         while(idx < interfaceIds.length) {
             setCollectionPrivate(interfaceIds[idx], instance, set);
+            idx += 1;
```
rudolfix: cool!

test/Universe.js
```diff
@@ -6,271 +6,542 @@ import registerSingletons from "./helpers/registerSingletons";
...
+        "platformTerms",
+      ];
+      for (let ii = 0; ii < kis.length; ii += 1) {
+        const loopi = await universe[getters[ii]]();
+        expect(loopi).to.eq(knownInstances[ii]);
+      }
+    });
 
-  it("should set singleton without convenience method");
+    it("should set singleton without convenience method");
```
rudolfix: the point of the test was to set a singleton that does not have convenience method (like universe.neumark()). a singleton where interface is just a random 4 bytes (not a knownInterface)

test/Universe.js
```diff
@@ -6,271 +6,542 @@ import registerSingletons from "./helpers/registerSingletons";
...
-      knownInterfaces.tokenExchangeRateOracle,
-      knownInterfaces.identityRegistry,
-      knownInterfaces.commitmentInterface,
-    ]);
-    // remove from collection
-    await universe.setCollectionInterface(knownInterfaces.commitmentInterface, addr, false, {
-      from: universeManager,
+
+    // i would say this is tested above
+    it("should remove interface from collection");
```
rudolfix: yeah please remove

huesforalice: done

test/Universe.js
huesforalice: done

huesforalice:
@rudolfix format fix is run, also I have reverted back to an older version of the yarn lock file, I hope we won't get those error messages any more

PR #137, opened 2018-08-16
add eto terms tests and some eto commitment tests

huesforalice:


test/ETO/ETOTerms.js
```diff
@@ -77,8 +78,165 @@ contract("ETOTerms", ([deployer, admin, investorDiscount, investorNoDiscount]) =
...
+      1,
+    );
+    terms.IS_CROWDFUNDING = true;
+    const termsValues = termsKeys.map(v => terms[v]);
+    // console.log(termsValues);
+    etoTerms = await ETOTerms.new.apply(this, termsValues);
+    await expect(etoTerms.requireValidTerms(platformTerms.address)).to.be.rejectedWith(EvmError);
+  });
+
+  it("should reject on platform terms with whitelist duration too small", async () => {
```
huesforalice: All the duration tests will also pass when setting the correct duration (here for example substitute the .sub(1) with sub(0) and it will be ok). It looks like I'm doing something wrong. @rudolfix could you take a look?

rudolfix: I know what is going on but it was tricky
```
// duration constraints
    uint256 public constant MIN_WHITELIST_DURATION_DAYS = 0 days;
    uint256 public constant MAX_WHITELIST_DURATION_DAYS = 60 days;
    uint256 public constant MIN_PUBLIC_DURATION_DAYS = 0 days;
    uint256 public constant MAX_PUBLIC_DURATION_DAYS = 60 days;

    // minimum length of whole offer
    uint256 public constant MIN_OFFER_DURATION_DAYS = 1 days;
    // quarter should be enough for everyone
    uint256 public constant MAX_OFFER_DURATION_DAYS = 90 days;
```

when you add maximum whitelist duration to public duration we cross 90 days of the offer. And check still fails. I talked to aga and we can modify platform terms so this is not the case:
maxium whitelist = 30
maximum public = 60
maximum total = 90

please modify platform terms and it should start working again

huesforalice:
@rudolfix I've added a new test for successfully setting the eto duration terms. Which fails. I think I'm somehow using the wrong. By changing the max duration the original problems also where not solved.

huesforalice:
Found the problem :)

PR #144, opened 2018-08-23
move token terms into own contract

huesforalice:


contracts/ETO/ETOTerms.sol
```diff
@@ -136,14 +130,12 @@ contract ETOTerms is Math {
...
         // test interface
         require(shareholderRights.HAS_GENERAL_INFORMATION_RIGHTS());
-        require(maxNumberOfTokens >= minNumberOfTokens);
+        require(tokenTerms.MAX_NUMBER_OF_TOKENS() >= tokenTerms.MIN_NUMBER_OF_TOKENS());
         require(shareNominalValueEurUlps > 0);
 
         DURATION_TERMS = durationTerms;
+        TOKEN_TERMS = tokenTerms;
         EXISTING_COMPANY_SHARES = existingCompanyShares;
-        MIN_NUMBER_OF_TOKENS = minNumberOfTokens;
```
rudolfix: what we should do: we should copy values that we use locally to storage and make them private. this way we save multiples of 900 gas units for inter-contract calls.

let's save gas!

rudolfix: `calculateContribution` is called for each investor in transaction context and they will pay for gas to do calculation. caching in local storage makes sense

PR #173, opened 2018-09-17
Added missing Universe tests with setting universe interface. Fixes #168

### LINKED ISSUE #168: Create missing test in Universe contrace

Exef:


test/Universe.js
```diff
@@ -562,6 +562,13 @@ contract(
       expect(cid[0]).to.eq(contractId("Universe"));
     });
 
-    it("should revert on setting KNOWN_INTERFACE_UNIVERSE singleton");
+    it("should revert on setting KNOWN_INTERFACE_UNIVERSE singleton", async () => {
+      await expect(universe.setSingleton(knownInterfaces.universe, universe.address)).to.revert;
```
rudolfix: this revert because it's not called from
```
 {
           from: universeManager,
         },
```
please add it to test. that's why revert codes are so important addition. you can add such revert code to the contract

PR #189, opened 2018-10-29
improves token controller tests

rudolfix:
1. adds many ETO Commitment tests - just a few are left
2. initial tests for equity token controller + test contracts
3. reorg of equity token tests


rudolfix:
we need to go through company governance in the token controller. we didn't do it at all. when back from devcon :>

PR #192, opened 2018-11-06
token code base unification & allowance override mechanism

rudolfix:
**single code base for all the tokens**
all remnants of zeppelin token were removed. now the internal structure of snapshot token code base is used everywhiere
1. BasicToken observes mTransfer and mOnTransfer internal interfaces methods (token + controller) exactly like BasicSnapshotToken
2. StandardToken removed entirely - now it's just is a mix of BasicToken and TokenAllowance which is also shared with StandardSnapshotToken

looks good to me!

**simplification of ITokenController**
onTransfer and onTransferFrom got merged in single function

**allowance override**
this mechanism is part of token controller (internal and external) and will allow
- forced transfers
- permanent allowances for gas exchange
- other non-trustless things for asset tokens

implementation of this call to token controller is optional in the token

also time has come to write general token controller test suite.


contracts/Company/PlaceholderEquityTokenController.sol
```diff
@@ -243,20 +243,12 @@ contract PlaceholderEquityTokenController is
...
-        return _transfersEnabled || from == _commitment;
-    }
-
-    function onTransferFrom(address, address, address, uint256)
-        public
-        constant
-        returns (bool allow)
-    {
-        return _transfersEnabled;
+        return _transfersEnabled || (from == _commitment && broker == from);
```
huesforalice: In general a short comment on what this should achieve is not so bad. To me it's clear that only the commitment should be able to make transfers if they're not globally enabled, but it just will help reading

contracts/SnapshotToken/Helpers/MTokenAllowanceController.sol
```diff
@@ -23,4 +23,20 @@ contract MTokenAllowanceController {
         internal
         returns (bool allow);
```
huesforalice: A question not really related to this PR, but why are some interfaces prefixed with M instead of I?

contracts/test/TestSnapshotToken.sol
```diff
@@ -79,6 +81,13 @@ contract TestSnapshotToken is
         _enableApprovals = enable;
     }
 
+    // this is really nasty controller that can force any transfer it wants
```
huesforalice: true anarchism :)

contracts/Company/PlaceholderEquityTokenController.sol
rudolfix: OK

contracts/SnapshotToken/Helpers/MTokenAllowanceController.sol
rudolfix: because this is internal interface which has internal functions, not public. and this is convention here

solidity has no `virtual` and `override` markers so it does not know if given method can be overriden and that you intend to override...

PR #193, opened 2018-11-06
Add test for blocking the EquityToken functionality using controller

### LINKED ISSUE #190: implement EquityToken tests against mocked token controller
tests in group `describe("IEquityTokenController tests", () => {`
1. Please add mock methods to `TestNullEquityTokenController` so you can control controller internal state
2. general idea is to test if controller can block and unblock all the token functions per interface
3. #189 must be reviewed and merged first

Exef:
Closes #190

### DIFF
--- a/contracts/test/ETO/TestNullEquityTokenController.sol
+++ b/contracts/test/ETO/TestNullEquityTokenController.sol
@@ -31,16 +31,6 @@ contract TestNullEquityTokenController is
     bool allowChangeNominee = true;
     bool allowChangeTokenController = true;
 
-    function resetAllowance()
-        public
-    {
-        allowOnTransfer = true;
-        allowDestroyTokens = true;
-        allowGenerateTokens = true;
-        allowChangeNominee = true;
-        allowChangeTokenController = true;
-    }
-
     //
     // Implements IEquityTokenController
     //
@@ -81,6 +71,10 @@ contract TestNullEquityTokenController is
         // msg.sender is ETOCommitment
     }
 
+    //
+    //  Implements ITokenController
+    //
+
     function onTransfer(address, address, uint256)
         public
         constant
@@ -88,13 +82,6 @@ contract TestNullEquityTokenController is
     {
         return allowOnTransfer;
     }
-
-    function setAllowOnTransfer(bool _allow)
-        public
-    {
-        allowOnTransfer = _allow;
-    }
-
 
     function onGenerateTokens(address, address, uint256)
         public
@@ -104,13 +91,6 @@ contract TestNullEquityTokenController is
         return allowGenerateTokens;
     }
 
-    function setAllowOnGenerateTokens(bool _allow)
-        public
-    {
-        allowGenerateTokens = _allow;
-    }
-
-
     function onDestroyTokens(address, address, uint256)
         public
         constant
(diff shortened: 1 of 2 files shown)


contracts/test/ETO/TestNullEquityTokenController.sol
```diff
@@ -25,6 +25,22 @@ contract TestNullEquityTokenController is
 
     }
 
+    bool allowOnTransfer = true;
```
rudolfix: please follow code formatting guidelines as found in https://github.com/Neufund/platform-contracts/blob/master/CodeStyle.md

Also look how sections are split in all contracts. We need strong conventions here. Mess is a way to lose money ;>

test/ETO/EquityToken.js
```diff
@@ -96,32 +99,113 @@ contract("EquityToken", ([admin, nominee, company, broker, ...holders]) => {
       expect(balance).to.be.bignumber.eq(initialBalance);
     });
 
-    it("should overflow on deposit");
+    it("should overflow on deposit", async () => {
+      const initialBalance = new web3.BigNumber(2).pow(256).plus(1);
```
rudolfix: you will not overflow this way....
 you need to do it in two steps

1. deposit 2**256-1 in first step - that should work
2. deposit 1 in next - that should revert
check final totalSupply and balance to be 2**256-1

test/ETO/EquityToken.js
```diff
@@ -96,32 +99,113 @@ contract("EquityToken", ([admin, nominee, company, broker, ...holders]) => {
...
+      const newEquityTokenController = await TestNullEquityTokenController.new(universe.address);
+
+      await equityToken.changeTokenController(newEquityTokenController.address);
+
+      expect(await equityToken.tokenController()).to.be.bignumber.eq(
+        newEquityTokenController.address,
+      );
+
+      // clean up
+      await equityToken.changeTokenController(equityTokenController.address);
```
rudolfix: you do not need to clean up after test. all tests start as clean slate - everything is redeployed.

test/ETO/EquityToken.js
```diff
@@ -96,32 +99,113 @@ contract("EquityToken", ([admin, nominee, company, broker, ...holders]) => {
...
+      await equityToken.changeNominee(newNominee, { from: company });
+
+      expect(await equityToken.nominee()).to.be.bignumber.eq(newNominee);
+
+      // clean up
+      await equityToken.changeNominee(nominee, { from: company });
+      expect(await equityToken.nominee()).to.be.bignumber.eq(nominee);
+    });
+
+    it("can block transfers", async () => {
```
rudolfix: 1. please also add tests for `transferFrom` and ERC223 transfers
2. add tests for approve() and `onApprove`
3. please add positive tests: can transfer, can approve etc.

why? because I want to use this code and write generic test suite for all tokens with controller so we need comprehensive test

test/ETO/EquityToken.js
Exef: That was a version that I had first.

contracts/test/ETO/TestNullEquityTokenController.sol
Exef: Improved.

test/ETO/EquityToken.js
```diff
@@ -96,32 +98,114 @@ contract("EquityToken", ([admin, nominee, company, broker, ...holders]) => {
...
 
-    it("should overflow on deposit");
+    it("should overflow on deposit", async () => {
+      const initialBalance = new web3.BigNumber(2).pow(256).minus(1);
+      await equityToken.issueTokens(initialBalance, {
+        from: company,
+      });
+
+      await expect(
+        equityToken.issueTokens(initialBalance, {
```
rudolfix: "1" should be enough to overflow now. please try with "1"

test/ETO/EquityToken.js
Exef: Ready. Please review once again.

PR #194, opened 2018-11-06
first version of disbursal contract

huesforalice:
not production ready but definitely dev and stage ready so we review and merge


contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,265 @@
...
+        require(isDisbursableToken(msg.sender), "NF_DISB_UKNOWN_TOKEN");
+        require(amount > 0, "NF_DISB_ZERO_AMOUNT");
+
+        // cast and check pro rata token
+        BasicSnapshotToken proRataToken = BasicSnapshotToken(decodeAddress(data));
+        uint256 snapshotId = proRataToken.currentSnapshotId() - 1;
+        require(proRataToken.totalSupplyAt(snapshotId) > 0, "");
+
+        // create a new disbursal entry
+        _disbursals[msg.sender].push(Disbursal({
```
huesforalice: Do I need to initialise this array somehow? Or can I just assume it's there and push to it?

rudolfix: no. you have some length prefixed structure already initialized

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    /// @notice claim a token, to be called an investor
+    /// @param token address of the recyclable token
+    /// @param investors list of investors we want to recycle tokens for
+    /// @param until until what index to claim to
+    function recycle(address token, address[] investors, uint256 until)
+    public
+    {        
+        // @TODO: Recycle funds
```
huesforalice: My main questions for this are:
- Do we create a new disbursal entry for every call of this function? Or do we accumlate the tokens and then disburse if there is a certain amount.
- Does the pro-rata token entry need to be retained, or do we default to NEU? If we have to retain the pro rata token, we can only do a redisbursal per original disbursal. 
- In any case, some tokens will be locked in limbo forever. Without saving which tokens have been claimed so far, we will never be able to get them back.

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+        // @TODO: add log message
+    }
+
+    // implementation of tokenfallback
+    function tokenFallback(address wallet, uint256 amount, bytes data)
+        public
+        only(ROLE_DISBURSER)
```
rudolfix: we do not need special role. delegate check to the controller.
msg.sender must be payment or equity/neu token (we check against collections in universe)
the wallet will be 1. ETOCommitment contract, 2. Equity Token Controller 3. anyone with ROLE_DISBURSER

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+        require(isDisbursableToken(msg.sender), "NF_DISB_UKNOWN_TOKEN");
+        require(amount > 0, "NF_DISB_ZERO_AMOUNT");
+
+        // cast and check pro rata token
+        BasicSnapshotToken proRataToken = BasicSnapshotToken(decodeAddress(data));
+        uint256 snapshotId = proRataToken.currentSnapshotId() - 1;
+        require(proRataToken.totalSupplyAt(snapshotId) > 0, "");
+
+        // create a new disbursal entry
+        _disbursals[msg.sender].push(Disbursal({
```
rudolfix: You should merge disbursal if they are made against same snapshotId. that's the whole point of having daily disbursal

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+        public
+        only(ROLE_DISBURSER)
+    {
+        // @TODO: move access control to Controller
+        require(isDisbursableToken(msg.sender), "NF_DISB_UKNOWN_TOKEN");
+        require(amount > 0, "NF_DISB_ZERO_AMOUNT");
+
+        // cast and check pro rata token
+        BasicSnapshotToken proRataToken = BasicSnapshotToken(decodeAddress(data));
+        uint256 snapshotId = proRataToken.currentSnapshotId() - 1;
```
rudolfix: OK. I thought about it. We should use proRataToken.currentSnapshotId() so the disbursal is "open" and we can merge more disbursals into it until the snapshot id advances. see below

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+            snapshotId: snapshotId,
+            disburser: wallet
+        }));
+        //@TODO: add log message
+    }
+
+    // implementation of reclaimbale
+    function reclaim(IBasicToken token)
+        public
+        only(ROLE_RECLAIMER)
```
rudolfix: should we really use it? if we cannot reclaim disbursed tokens then this is not really helpful. the typical problem will be people sending tokens to the pool with normal `transfer` function but those probably will be euro and ether tokens that can be reclaimed. let's discuss it. anyway you do not need to do `only` here. it will be done in the base class

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    /// @notice helper to determine if the token at the given address is supported for disbursing
+    /// @param token address of token in question
+    function isDisbursableToken(address token)
+        public
+        view
+        returns (bool)
+    {   
+        // @TODO: migrate this to new, more flexible token registering in universe
```
rudolfix: move to controller
those could be payment tokens and asset tokens via respective collections
we'll also put NEU in the asset tokens

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+
+    /// @notice claim a token for an spender, returns the amount of tokens claimed
+    /// @param token address of the claimable token
+    /// @param spender address of the spender that will receive the funds
+    /// @param until until what index to claim to
+    function claimPrivate(address token, address spender, uint256 until)
+    public
+    returns (uint256 claimedAmount, uint256 lastIndex)
+    {
+        require(isDisbursableToken(token), "NF_DISB_UKNOWN_TOKEN");
```
rudolfix: permissions via controller

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+        //@TODO: add log message
+    }
+
+    /// @notice get the amount of tokens that can be claimed by a given spender
+    /// @param token address of the claimable token
+    /// @param spender address of the spender that will receive the funds
+    /// @param until until what index to claim to, use UINT256_MAX for all
+    function claimablePrivate(address token, address spender, uint256 until)
+    public
+    view
```
rudolfix: let's use `constant`. that's the convention in the code. at least for now.

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+    returns (uint256 claimableAmount, uint256 lastIndex)
+    {
+        require(isDisbursableToken(token), "NF_DISB_UKNOWN_TOKEN");
+        lastIndex = min(until, _disbursals[token].length);
+        claimableAmount = 0;
+        for (uint256 i = _disbursalProgress[token][spender]; i < lastIndex; i += 1) {
+            Disbursal storage disbursal = _disbursals[token][i];
+            BasicSnapshotToken proRataToken = BasicSnapshotToken(disbursal.proRataToken);
+            uint256 snapshotId = disbursal.snapshotId;
+            uint256 proRataTotalSupply = proRataToken.totalSupplyAt(snapshotId);
```
rudolfix: please store this in Disbursal struct! call is 900 unit of gas then you need to access storage in internal struct of the token. so it's better to cache it. it's 200 gas units per slot read. and we are shaving off gas.
so it makes sense to cache it for the good of the community. first claimer would pay for this obviously but we can assume that he's going to be a rich guy

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+            uint256 proRataSpenderBalance = proRataToken.balanceOfAt(spender, snapshotId);
+            if (proRataTotalSupply == 0) continue;
+            // @TODO: do we need checks for overflow here? probably not as uint256 is really large..
+            // this should round down, so we should not be spending more than we have in our balance
+            claimableAmount += (proRataSpenderBalance * disbursal.amount) / proRataTotalSupply;
```
rudolfix: please use our fully tested `Math` contract. it has `proportion` method that does exactly what you have here but for example makes sure there are no overflows and rounding is consistent.

also. we need a real proof that rounding errors here do not accumulate. if they do, the last person may not be able to claim because there is not enough balance.

but I have inductive proof that it will not happen. it's however not so obvious at all!

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+    /// @notice get the amount of tokens that can be claimed by a given spender
+    /// @param token address of the claimable token
+    /// @param spender address of the spender that will receive the funds
+    /// @param until until what index to claim to, use UINT256_MAX for all
+    function claimablePrivate(address token, address spender, uint256 until)
+    public
+    view
+    returns (uint256 claimableAmount, uint256 lastIndex)
+    {
+        require(isDisbursableToken(token), "NF_DISB_UKNOWN_TOKEN");
```
rudolfix: here we should prevent reclaiming if `currentSnapshotId` is identical to snapshot in Disbursal struct. because for current snapshot id, balances and total supply are not sealed

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+        // snapshop ID of the pro-rata token, which will define which amounts to disburse against
+        uint256 snapshotId;
+        // amount of tokens to disburse
+        uint256 amount;
+        // address of the token used to determine the user pro rata amount, must be a snapshottoken, default is NEU
+        BasicSnapshotToken proRataToken;
+        // time after which claims to this token can be recycled
+        uint256 recycleAfterTimestamp;
+        // contract sending the disbursal
+        address disburser;
```
rudolfix: ok. makes sense to have it here

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+contract FeeDisbursal is
+    AccessControlled,
+    Reclaimable,
+    IFeeDisbursal,
+    IdentityRecord,
+    Serialization,
+    Math
+{
+
+    uint256 constant UINT256_MAX = 2**256 - 1;
```
rudolfix: make section for constants. I think it goes after types ;>

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    ////////////////////////
+    // Public functions
+    ////////////////////////
+
+    /// @notice get the disbursal at a given index for a given token
+    /// @param token address of the claimable token
+    /// @param index until what index to claim to
+    function getDisbursal(address token, uint256 index)
```
rudolfix: If you make iterator then also provide length somewhere. Or we iterate until we have 0 return struct?

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+    public
+    returns (uint256[])
+    {
+        // only allow verified and active accounts to claim tokens
+        // @TODO: move access control to Controller
+        IdentityClaims memory claims = deserializeClaims(IDENTITY_REGISTRY.getClaims(msg.sender));
+        require(claims.isVerified && !claims.accountFrozen, "NF_DISB_NOT_VER");
+        uint256[] memory result = new uint256[](tokens.length);
+        for (uint256 i = 0; i < tokens.length; i += 1) {
+            (uint256 claimedAmount, uint256 lastIndex)  = claimPrivate(tokens[i], msg.sender, UINT256_MAX);
```
rudolfix: it may be a little bit too much but it's a good practice to make all state changes first and then call external contract
1. make a bookkeeping in one loop
2. make transfer calls in the another loop

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+            result[i] = claimedAmount;
+        }
+        return result;
```
rudolfix: please note that external clients do not have access to return value of non-constant function. this is only available for intra contract calls.
also maybe this function should be `external`
or maybe you plan to call it with `eth_call` as a kind of dry run? then it makes sense

contracts/FeeDisbursal/FeeDisbursal.sol
rudolfix: ad 1. if you implement merging latest disbursals with current snapshot id then this problem is gone
ad 2. hmmm I do not get it. you do disbursal per token originally made.
ad 3. why? you can recover past balances of the shareToken to get who didn't claim yet... not easy but can be done.

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: done in all 5 cases

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: Done, good point!

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: Replacing the proportion part is done

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: :) Done

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: I don't get it exactly. Do you want me to split up the claimPrivate function into two others? In that core function it#s done exactly like this. first update state, then call transfer.

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: No I was not planning that. Thanks for pointing this out, I will remove all the returns for non constant functions

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: yeah, thanks, works like this :)

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: I will kick it out for now. I thought it was good practice to have some kind of fallback

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: added a count function

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: ok!

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,269 @@
...
+    ////////////////////////
+    // Constructor
+    ////////////////////////
+    constructor(Universe universe)
+        AccessControlled(universe.accessPolicy())
+        Reclaimable()
+        public
+    {
+        UNIVERSE = universe;
+        IDENTITY_REGISTRY = IIdentityRegistry(universe.identityRegistry());
```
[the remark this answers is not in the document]
huesforalice: done

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: done

rudolfix:
@huesforalice can you give more descriptive commit messages? every commit should say what it does

huesforalice:
@rudolfix will do

contracts/FeeDisbursal/FeeDisbursal.sol
huesforalice: Actually I though about it. If we have unsealed disbursals, storing this will not work properly, as we might have a mismatch of totalSupply at the point of claiming, because there could always be more prorata tokens minted after the disbursal was made, no?

contracts/FeeDisbursal/FeeDisbursalController.sol
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    function onDisburse(address token, address disburser, uint256 amount, address /*proRataToken*/)
+        public
+        returns (bool allow)
+    {   
+        //@TODO: should we dissalow token and pro rata token to be the same?
+        bool disburserAllowed = 
+            UNIVERSE.isAnyOfInterfaceCollectionInstance(DISBURSE_ALLOWED_INTERFACES, disburser) ||
+            UNIVERSE.isSingleton(KNOWN_INTERFACE_FEE_DISBURSAL, disburser) ||
```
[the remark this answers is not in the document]
huesforalice: Yep, that's right.

contracts/FeeDisbursal/FeeDisbursalController.sol
```diff
@@ -0,0 +1,108 @@
...
+        IdentityClaims memory claims = deserializeClaims(IDENTITY_REGISTRY.getClaims(spender));
+        return isDisbursableToken(token) && claims.isVerified && !claims.accountFrozen;
+    }
+
+    function onDisburse(address token, address disburser, uint256 amount, address /*proRataToken*/)
+        public
+        returns (bool allow)
+    {   
+        //@TODO: should we dissalow token and pro rata token to be the same?
+        bool disburserAllowed =
```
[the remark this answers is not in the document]
huesforalice: ok, done

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,378 @@
...
+    returns (uint256 claimedAmount, uint256 lastIndex)
+    {
+        (claimedAmount, lastIndex) = claimablePrivate(token, spender, until, false);
+
+        // mark spender disbursal progress
+        _disbursalProgress[token][spender] = lastIndex;
+
+        // do the actual token transfer
+        IERC223Token ierc223Token = IERC223Token(token);
+        ierc223Token.transfer(spender, claimedAmount, "");
```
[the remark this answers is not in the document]
huesforalice: No it's not...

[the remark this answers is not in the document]
huesforalice: Done, though

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,378 @@
...
+        // also move the _disbursalProgress pointer
+        uint256 totalAmount = 0;
+        for (uint256 i = 0; i < investors.length; i += 1) {
+            (uint256 claimableAmount, uint256 lastIndex) = claimablePrivate(token, investors[i], until, true);
+            totalAmount += claimableAmount;
+            _disbursalProgress[token][investors[i]] = lastIndex;
+        }
+
+        // now re-disburse, we're now the disburser
+        disburse(token, this, totalAmount, UNIVERSE.neumark());
```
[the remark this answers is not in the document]
huesforalice: probably default recycle time :)

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,378 @@
...
+
+        uint256 snapshotId = proRataToken.currentSnapshotId();
+        uint256 proRataTokenTotalSupply = proRataToken.totalSupplyAt(snapshotId);
+        require(proRataTokenTotalSupply > 0, "");
+
+        Disbursal[] storage disbursals = _disbursals[token];
+
+        // try to merge with an existing disbursal
+        bool merged = false;
+        for ( uint256 i = disbursals.length - 1; i < disbursals.length; i-- ) {
```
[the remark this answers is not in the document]
huesforalice: i>=0 will always be true on an unsigned int. I'm now explictely checking against UINT_MAX to show how it should work. Open to other suggestions though. And why does linter not like i--? This is the only way for me to do this, I have to go through the array backwards.

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,378 @@
...
+    /// @param amount amount of the disbursable tokens
+    /// @param proRataToken address of the token that defines the pro rata
+    function disburse(address token, address disburser, uint256 amount, BasicSnapshotToken proRataToken)
+    internal
+    {
+        require(_feeDisbursalController.onDisburse(token, disburser, amount, address(proRataToken)), "");
+
+        uint256 snapshotId = proRataToken.currentSnapshotId();
+        uint256 proRataTokenTotalSupply = proRataToken.totalSupplyAt(snapshotId);
+        require(proRataTokenTotalSupply > 0, "");
```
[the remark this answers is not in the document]
huesforalice: done

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,378 @@
...
+                disbursal.disburser == disburser) {
+                merged = true;
+                disbursal.amount += amount;
+            }
+        }
+
+        // create a new disbursal entry
+        if (!merged) 
+            disbursals.push(Disbursal({
+                recycleableAfterTimestamp: block.timestamp + 365 days,
```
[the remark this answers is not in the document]
huesforalice: Done (not with a struct but with a second packed argument).

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,378 @@
...
+
+        Disbursal[] storage disbursals = _disbursals[token];
+
+        // try to merge with an existing disbursal
+        bool merged = false;
+        for ( uint256 i = disbursals.length - 1; i < disbursals.length; i-- ) {
+            // we can only merge if we have the same snapshot id
+            // we can break here, as continuing down the loop the snapshot ids will decrease
+            Disbursal storage disbursal = disbursals[i];
+            if ( disbursal.snapshotId < snapshotId )
```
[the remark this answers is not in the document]
huesforalice: That does not make sense to me. If we merge disbursals that have different pro rata tokens set, this will end in total chaos. I can't think of the case where this would work, but maybe we can talk about it to be clear.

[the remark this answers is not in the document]
rudolfix: I do not agree. Let's discuss this

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -0,0 +1,387 @@
...
+        emit LogChangeFeeDisbursalController(oldController, address(newController), msg.sender);
+    }
+
+    /// @notice implementation of tokenfallback, calls the internal disburse function
+    /// legacy onTokenTransfer is also supported via imported file
+    function tokenFallback(address wallet, uint256 amount, bytes data)
+        public
+    {
+        uint256 recycleAfterPeriod;
+        BasicSnapshotToken proRataToken;
```
rudolfix: IMO we should take all defaults from disbursal controller. up to be discussed

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -109,129 +120,201 @@ contract FeeDisbursal is
...
-    function claimMultiple(address[] tokens)
-    public
+    /// @param proRataToken address of the token used to determine the user pro rata amount, must be a snapshottoken
+    function acceptMultiple(address[] tokens, ITokenSnapshots proRataToken)
+        public
+    {
+        uint256[2][] memory claimed = new uint256[2][](tokens.length);
+        // first gather the funds
+        uint256 i;
+        uint256 totalAmount;
```
huesforalice: We don't really need this, do we?

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -288,100 +372,111 @@ contract FeeDisbursal is
...
+            if ( snapshotId == currentSnapshotId )
                 break;
-            uint256 proRataTokenTotalSupply = proRataToken.totalSupplyAt(snapshotId);
-            uint256 proRataSpenderBalance = proRataToken.balanceOfAt(spender, snapshotId);
-            if (proRataTokenTotalSupply == 0 || proRataSpenderBalance == 0) continue;
-            // this should round down, so we should not be spending more than we have in our balance
-            claimableAmount += proportion(disbursal.amount, proRataSpenderBalance, proRataTokenTotalSupply);
+            // in case of just determining the recyclable amount of tokens, break when we
+            // cross this time, this also assumes disbursal.recycleableAfterTimestamp in each disbursal is the same or increases
+            // in case it decreases, recycle will not be possible until 'blocking' disbursal also expires
```
huesforalice: what exactly does this comment mean?

contracts/FeeDisbursal/FeeDisbursal.sol
rudolfix: according to designs we still need it. it will let you take all funds from single pro rata token (like NEUMARK) in one transaction. to me it makes sense. we still miss a method where you can take funds for single payment token but for multiple proRataTokens. then we save a lot of gas on transfers. I still plan to add it.

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -288,100 +372,111 @@ contract FeeDisbursal is
     /// @param disburser address of the actor disbursing (e.g. eto commitment)
     /// @param amount amount of the disbursable tokens
     /// @param proRataToken address of the token that defines the pro rata
```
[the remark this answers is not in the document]
rudolfix: will update

contracts/FeeDisbursal/FeeDisbursal.sol
rudolfix: if you do disbursal with short time to recycle then with long time and then another with short time, you will not be able to recycle 1 and 3 when time passes because disbursal 2 will not yet expire and you will not reach 3. this is how disbursalProgress works. I'll write test for it

contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -358,6 +434,10 @@ contract FeeDisbursal is
 
         uint256 snapshotId = proRataToken.currentSnapshotId();
```
huesforalice: I would like to briefly talk to you about how this works.

PR #206, opened 2018-12-06
Fixed har-validator dependency: 5.1.2 was removed by publisher

usetech-llc:
Using 5.1.3 instead (deleted yarn.lock and ran yarn install)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,11 +2,6 @@
 # yarn lockfile v1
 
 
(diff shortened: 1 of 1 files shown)


rudolfix:
@usetech-llc hey! we alrady fixed this in the current master. however tests are not passing on this PR which makes me a little worried. we'll investigate a little bit what happened. maybe new ganache version?

PR #222, opened 2019-02-09
completes fee disbursal tests

rudolfix:


contracts/FeeDisbursal/FeeDisbursal.sol
```diff
@@ -599,7 +613,11 @@ contract FeeDisbursal is
...
             proRataTokenTotalSupply -= proRataToken.balanceOfAt(address(this), snapshotId);
         }
-        // TODO: prove that rounding errors do not accumulate here
-        return proportion(disbursalAmount, proRataClaimerBalance, proRataTokenTotalSupply);
+        // using round HALF_UP we risks rounding errors to accumulate and overflow balance at the last claimer
+        // example: disbursalAmount = 3, total supply = 2 and two claimers with 1 pro rata token balance
+        // with HALF_UP first claims 2 and seconds claims2 but balance is 1 at that point
+        // thus we round down here saving tons of gas by not doing additional bookkeeping
+        // consequence: small amounts of disbursed funds will be left in the contract
+        return mul(disbursalAmount, proRataClaimerBalance) / proRataTokenTotalSupply;
```
huesforalice: This looks good to me. I thought there was a way to prove half up somehow added up in the end. But if we can't do it, this is fine

rudolfix: actually you can prove otherwise with negative example as above

PR #238, opened 2019-03-25
Added new investors fixtures

Exef:


migrations/config.js
```diff
@@ -143,11 +143,31 @@ export function getFixtureAccounts(accounts) {
       "investor",
       false,
     ),
+
+    INV_ETH_ICBM_NO_KYC_2: makeAccount(
+      "0xDE185A5c2Bd3913fAC1F64102e3DEFD9E1797C4d",
+      "investor",
+      true,
```
rudolfix: I assume there should be `false`? that's no KYC account

PR #254, opened 2019-05-14
Seed, PK and derivation path in the fixture configuration

Exef:


migrations/101_apply_fixtures.js
```diff
@@ -166,7 +166,7 @@ module.exports = function deployContracts(deployer, network, accounts) {
     await icbmEuroToken.setAllowedTransferTo(euroLock.address, true);
 
     console.log("migrating locked accounts");
-    await icbmEtherLock.migrate({ from: fas.INV_ETH_EUR_ICBM_M_HAS_KYC.address });
```
rudolfix: AFAIK `INV_ETH_EUR_ICBM_M_HAS_KYC` was the one to be migrated, why you changed this single fixture?

PR #255, opened 2019-05-15
Multi wallet provider

rudolfix:
1. make multi wallet provider actually work
2. use it as provider for following network in truffle.js: localhost, nf_private, nf_private_io
3. make sure fixture accounts have sufficient ether before you use them in migration scripts
4. make sure DEPLOYER (`0x8a194c13308326173423119f8dcb785ce14c732b`) is in the fixture accounts
5. you should generate artifacts and test them with backend by copying them and running all the tests


migrations/101_distrubute_ether_to_fixtures_accounts.js
```diff
@@ -0,0 +1,29 @@
+require("babel-register");
+const getConfig = require("./config").getConfig;
+const getFixtureAccounts = require("./config").getFixtureAccounts;
+const getDeployerAccount = require("./config").getDeployerAccount;
+const promisify = require("../test/helpers/evmCommands").promisify;
+
+module.exports = function deployContracts(deployer, network, accounts) {
```
rudolfix: you should do it before  `migrations/7_patch_icbm_state.js`. could be even in first step which is empty now

migrations/102_patch_icbm_state.js
```diff
@@ -62,11 +62,12 @@ module.exports = function deployContracts(deployer, network, accounts) {
       from: fas.INV_EUR_ICBM_HAS_KYC.address,
     });
     await commitment.commitEuro({ from: fas.INV_EUR_ICBM_HAS_KYC.address });
-
     await euroToken.deposit(fas.INV_EUR_ICBM_HAS_KYC_2.address, amountEur);
+
```
rudolfix: you need to keep it number 7

multiWalletProvider.js
```diff
@@ -0,0 +1,65 @@
...
+
+export function multiWalletProvider(nodeUrl) {
+  const web3HttpProvider = new Web3.providers.HttpProvider(nodeUrl);
+  const engine = new ProviderEngine();
+
+  const fas = getFixtureAccounts();
+  const wallets = [];
+  for (const name of Object.keys(fas)) {
+    // TODO: Remove this check
+    if (fas[name].privateKey !== null) {
```
rudolfix: why? it's good. we can have external accounts that do not have private keys. like NANO LEDGERs. still we can send them ether or NEU

PR #260, opened 2019-05-20
adds constraints to eto mapping into eto fixtures

huesforalice:


migrations/configETOTermsFixtures.js
```diff
@@ -128,3 +128,14 @@ export const constraints = [
 
 // this will be populated in migration step 12 and then can be used when deploying mock ETOs
 export const deployedAddresses = [];
+
+export const getFixtureAndAddressByName = name => {
+  for (let i = 0; i < constraints.length; i++) {
```
tl-neufund: You could probably do something like this here instead of the for loop but it's just a stylistic thing:

```javascript
return constraints.find(constraint => constraint.name == name).map(constraint => ({
        constraintFixture: constraints[i],
        constraintAddress: deployedAddresses[i],
}))
```

PR #261, opened 2019-05-20
remove int max from terms, terms constraints and eto commitment

huesforalice:


test/ETO/ETOCommitment.js
```diff
@@ -491,6 +491,17 @@ contract("ETOCommitment", ([deployer, admin, company, nominee, ...investors]) =>
       );
     });
 
+    it("interprets MAX_TICKET_EUR_ULPS == 0 as unlimited", async () => {
```
huesforalice: This test is now modified and the title does not make sense anymore. MAX_TICKET_EUR_ULPS is not set to 0.

PR #266, opened 2019-05-23
Initial ETH amount and investment for fixture accounts.

Exef:


migrations/106_deploy_fixture_commitments.js
```diff
@@ -21,6 +21,139 @@ const Q18 = require("../test/helpers/constants").Q18;
 const CommitmentState = require("../test/helpers/commitmentState").CommitmentState;
 const toChecksumAddress = require("web3-utils").toChecksumAddress;
 
+function getWhitelistForETO(etoDefinition, fas) {
+  console.log(etoDefinition.name);
+  const whitelist = [];
+
+  for (const f of Object.keys(fas)) {
```
tl-neufund: This code isn't tested but it's worth using javascript features rather than for loops if you can because it's more readable, but no worries we can still merge like this

```javascript
const whitelist = Object.entries(fas).reduce(([fKey, fObj], whitelist) => {
    const {etoParticipation: {whitelist: thisEtoWhitelist} } = fObj;
   
    if(!thisEtoWhitelist) {return whitelist};

    const whitelistObjs = Object.entries()
      .find(([etoName, whitelistInfo]) => etoName == etoDefinition.name)
      .map(([etoName, whitelistInfo]) => {
        address: fObj.address,
        discountAmount: whitelistInfo.discountAmount,
        discount: whitelistInfo,
      })
    whitelist.append()
  }
}, [])
```

PR #282, opened 2019-07-01
assigns whitelist manager role to eto

rudolfix:
it's a first step to make whitelist syncing service that will require that role


contracts/Company/ShareholderRights.sol
```diff
@@ -30,6 +30,11 @@ contract ShareholderRights is IContractId {
     // Immutable state
     ////////////////////////
 
+    // todo: split into ShareholderRights and TokenholderRigths where the first one corresponds to rights of real shareholder (nominee, founder)
```
Exef: Please, create the corresponding issue

PR #299, opened 2019-09-03
adds contract deployment via api

rudolfix:


migrations/deployETO.js
```diff
@@ -266,7 +272,7 @@ export async function checkETO(artifacts, config, etoCommitmentAddress) {
   );
   const termsInUniverse = await universe.isInterfaceCollectionInstance(
     knownInterfaces.termsInterface,
-    await eto.etoTerms(),
+    await etoTerms.address,
```
desfero: seems like we don't need to `await` here?

rudolfix: yeah! will fix thx

PR #308, opened 2019-10-03
improves deployment toolset

rudolfix:
commit list id pretty self explanatory regarding scope


huesforalice:
Approved, wondering if changing the values in the MockEtoConstraints might break integration tests on the backend though?

rudolfix:
@huesforalice I can only say: we'll see but this changes some lower limits which we do not actually use when deploying fixtures

PR #310, opened 2019-10-08
snapshotted Voting Contract + tests

djudjuu:
voting contract and test cases as specified [here](https://github.com/Neufund/platform-contracts/blob/voting-task/contracts/SnapshotToken/Extensions/VOTING.md)

Current Status:
- [x] everybody can propose
- [x] results based on quorum & majority thresholds
- [x] snapshot based voting
- [x] testsuite integration for all of the above 
- [x] relayed voting
- [x] batched relayed voting

### Clarifications

> "the voting quorum is X% of all tokens and voting majority is > 50% of voting quorum" 

   I read this as: >X% of all tokens need to be involved in a vote to make it
   legitimate. For any legitimate vote, more than 50% of the involved tokens
   need to be in favor for a proposal to pass.

> bind snapshot token instance **ITokenSnapshots** in constructor

   this Interface only specifies the "currentSnapshotId()"-method, which will give no guarantees, that the balances at that snapshot are no longer subject to change once the vote is created. To guarantee that the token needs to implement the method "createSnapshot()" or "advanceSnapshotId()", whereas the latter not always guarantees that the snapshotID will actually be advanced (which i think is potentially dangerous, but could be discussed elsewhere).

So I choose to require the token to inherit from ISnapshotableToken, going for  "createSnapshot()", (as also done by Disbursal.sol)

### On relaying architecture

To relay a vote the *voter* has to compose and hash a message (see below for details), sign it and then pass the signature, together with the indended voting paramters (proposalIndex + bool) to a relayer willing to pay for the gas. The relayer then calls the ```relayedVote()``` function on the contract and passes in the voting parameters and the (decomposed) signature. The constract reconstructs the signed message from the voting parameters and uses ecrecover to verify that the signer is indeed the *voter*.

The implemented solution does not reward the relayer in any way (not part of the spec). It is simple in that the only data to be transmitted is a signature. If the relayer has collected votes from multiple voters, those can be batched together and all executed with a single transaction by calling ```batchRelayedVotes()```.

The message for the relayer is constructed by hashing the following pieces of data:
-  "EthereumSignedMessage\n"
- 32 (length of the message)
-  "0x00" to signal "Data with intended validator" according to [EIP 191](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-191.md) 
- the address of the voting contract, aka the intended validator
- the index of the proposal to vote on
- a boolean to signal whether to accept or reject the proposal

### Assumptions / Things to discuss

- 1) **data structure:** I decided for the ```_proposal``` to be a mapping so that we can delete Proposaldata after the end of the voting and not have to worry about using indices twice. If we wanted to keep Proposaldata around (e.g. for querying by other contracts) a simple array would be more appropriate
- 2) The **general composition of the vote functions**, whether they return something and what is logged in events would be something to go over if we wanted to use this further. For now, I assumed that there is an offchain registry handling the contents of the proposals, so that having an index for them is enough. Also, events have the minimal information to deduce info such as necessary amount of tokens to vote. I assume the quorum is known. Atm, there is no way for other contracts to act on a proposal-outcome. If we wanted to enable that it would probably be best to save the outcome of a proposal on its datastructure and put in a function to query that.
- 3)  Atm, I **assume** the totalSupply of the Token will not be maxed out and I simply multiply it with the Quorum without checking for overflows. See [comment below](https://github.com/Neufund/platform-contracts/pull/310#discussion_r332446783), Instead, we could implement overflow-checks, inherit from Math.sol or use a deployed Library. Typically, this would be something I'd want to discuss. (My opinion would be to inherit from Math.sol)

### Questions (not urgent though)

I don't understand why we are not losing information the way that setValue() is implemented [here[(https://github.com/djudjuu/platform-contracts/blob/6e04206532d9ff3a6511252e977edb5f3fb0d4b9/contracts/Snapshot/Snapshot.sol#L165).  I tried to break it though and couldnt, so i guess it is working fine.. Still , I can't wrap my head around it, though. One thing that confused me was the, I find, ill-named boolean at the beginning of the if-else clause (```hasNewSnapshot```. See the feedback-section of my [worklog](https://gist.github.com/djudjuu/5fbce7b8fc2218768531186394476752) for an elaboration.
 
=====================
working log can be found [here](https://gist.github.com/djudjuu/5fbce7b8fc2218768531186394476752)

### DIFF
--- a/contracts/SnapshotToken/Extensions/SimpleVote.sol
+++ b/contracts/SnapshotToken/Extensions/SimpleVote.sol
@@ -82,16 +82,69 @@ contract SimpleVote {
 
     }
 
-    /// @dev weighed vote on a given proposal, throws if proposal does not exist or the vote on it
-    ///   has ended already. Votes are final, changing the vote is not allowed
+    /// @dev increase the votecount on a given proposal by the token balance of the sender
+    ///   throws if proposal does not exist or the vote on it has ended already. Votes are final,
+    ///   changing the vote is not allowed
     /// @param proposalIndex of the proposal to be voted on
     /// @param voteInFavor of the desired proposal
     function vote(uint256 proposalIndex, bool voteInFavor) public {
+        incrementVote(proposalIndex, voteInFavor, msg.sender);
+    }
+
+    /// @dev same as vote, only for a relayed vote. Will throw if provided signature (v,r,s) does not match
+    ///  the address of the voter
+    /// @param voter address whose token balance should be used as voting weight
+    function relayedVote(
+        uint256 proposalIndex,
+        bool voteInFavor,
+        address voter,
+        bytes32 r,
+        bytes32 s,
+        uint8 v
+    )
+        public
+    {
+        // check that message signature matches the voter address
+        // solium-disable indentation
+        require(ecrecover(
+            keccak256(abi.encodePacked(
+                "\x19Ethereum Signed Message:\n32",
+                keccak256(abi.encodePacked(byte(0), this, proposalIndex, voteInFavor)))),
+            v, r, s) == voter,
+        "incorrect order signature");
+        // solium-enable indentation
+
+        incrementVote(proposalIndex, voteInFavor, voter);
+    }
+
+    /// @return whether the vote on a proposal has passed, if false this could be because it was rejected or
+    ///   because the quorum was not reached.
+    /// @dev throws if vote is still ongoing or proposal does not exist
+    function getResult(uint256 proposalIndex) public {
+        require(proposalIndex < _proposals.length, "Invalid proposal index");
+        Proposal storage p = _proposals[proposalIndex];
+        require(p.endTime < now, "Voting period is not over");
+
+        uint256 minParticipation = TOKEN.totalSupplyAt(p.snapshotId) * 40 / 100;
+        bool passedQuorum = p.inFavor + p.against > minParticipation;
+        bool hasPassed = passedQuorum && p.inFavor > p.against;
+        _hasActiveProposal[_proposals[proposalIndex].owner] = false;
+
+        emit LogProposalResult(proposalIndex, hasPassed, p.inFavor, p.against);
+    }
+
+    /// @dev increase the votecount on a given proposal by the token balance of a given address,
+    ///   throws if proposal does not exist or the vote on it has ended already. Votes are final,
+    ///   changing the vote is not allowed
+    /// @param proposalIndex of the proposal to be voted on
+    /// @param voteInFavor of the desired proposal
+    /// @param voter address whose tokenBalance is to be used as voting-weight
+    function incrementVote(uint256 proposalIndex, bool voteInFavor, address voter) internal {
         require(proposalIndex < _proposals.length, "Invalid proposal index");
         Proposal storage p = _proposals[proposalIndex];
         require(p.endTime > now, "Voting period is over");
-        require(!p.hasVoted[msg.sender]);
-        uint256 weight = TOKEN.balanceOfAt(msg.sender, p.snapshotId);
+        require(!p.hasVoted[voter]);
+        uint256 weight = TOKEN.balanceOfAt(voter, p.snapshotId);
         require(weight > 0, "Token balance at proposal time is zero");
 
         if (voteInFavor) {
@@ -99,21 +152,8 @@ contract SimpleVote {
         } else {
             p.against += weight;
         }
-        p.hasVoted[msg.sender] = true;
+        p.hasVoted[voter] = true;
     }
 
-    /// @return whether the vote on a proposal has passed, if false this could be because it was rejected or
-    ///   because the quorum was not reached.
-    /// @dev throws if vote is still ongoing or proposal does not exist
-    function getResult(uint256 proposalIndex) public returns(bool hasPassed) {
-        require(proposalIndex < _proposals.length, "Invalid proposal index");
-        Proposal storage p = _proposals[proposalIndex];
-        require(p.endTime < now, "Voting period is not over");
 
-        uint256 minParticipation = TOKEN.totalSupplyAt(p.snapshotId) * 40 / 100;
-        bool passedQuorum = p.inFavor + p.against > minParticipation;
-        hasPassed = passedQuorum && p.inFavor > p.against;
-        _hasActiveProposal[_proposals[proposalIndex].owner] = false;
-        emit LogProposalResult(proposalIndex, hasPassed, p.inFavor, p.against);
-    }
 }
--- a/contracts/SnapshotToken/Extensions/SimpleVote.sol
+++ b/contracts/SnapshotToken/Extensions/SimpleVote.sol
@@ -3,12 +3,12 @@ pragma solidity 0.4.26;
 import "./ISnapshotableToken.sol";
 /* import "../../Math.sol"; */
 
+/// Contract to allow weighted voting based on a snapshotable token (with relayed, batched voting)
 contract SimpleVote {
 
     ////////////////////////
     // Types
     ////////////////////////
-
     struct Proposal {
         uint256 inFavor;
         uint256 against;
@@ -16,35 +16,35 @@ contract SimpleVote {
         address owner;
         uint256 snapshotId;
         mapping (address => bool) hasVoted;
-
     }
 
     /////////////////////////
     // Immutable state
     ////////////////////////
-    uint256 private VOTING_PERIOD;
-    ISnapshotableToken private TOKEN;
+    ISnapshotableToken public TOKEN;
     // % of tokens to participate in a voting to make it valid
-    uint256 QUORUM = 40;
+    uint256 public QUORUM = 40;
+    uint256 public VOTING_PERIOD;
 
     /////////////////////////
     // Mutable state
     ////////////////////////
-    Proposal[] private _proposals;
+    mapping (uint256 => Proposal) private _proposals;
     mapping (address => bool) private _hasActiveProposal;
+    uint256 private _nextProposalId;
 
     /////////////////////////
     // Events
     ////////////////////////
-    event LogNewProposal(address indexed owner, uint256 indexed proposalIndex, uint256 snapshotId, uint256 endTime);
-    event LogProposalResult(uint256 indexed proposalIndex, bool hasPassed, uint256 inFavor, uint256 against);
+    event LogNewProposal(address indexed owner, uint256 indexed proposalId, uint256 snapshotId, uint256 endTime);
+    event LogProposalResult(uint256 indexed proposalId, bool hasPassed, uint256 inFavor, uint256 against);
 
     ////////////////////////
     // Constructor
     ////////////////////////
 
-    /// @param votingPeriodInDays amount of days that each vote will take
     /// @param token which balances are used for voting
+    /// @param votingPeriodInDays amount of days that each vote will take
     constructor(
         ISnapshotableToken token,
         uint256 votingPeriodInDays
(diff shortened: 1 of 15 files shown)


contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,119 @@
...
+
+    /////////////////////////
+    // Public functions
+    ////////////////////////
+
+    /// @dev Creates a proposal that will end VOTING_PERIOD days from now
+    function addProposal() public {
+        require(!_hasActiveProposal[msg.sender]);
+
+        uint256 sId = TOKEN.currentSnapshotId();
```
djudjuu: the comment for createSnapshot() said not to rely on the returned integer, so I am doing two separate calls.

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,119 @@
...
+
+    /// @return whether the vote on a proposal has passed, if false this could be because it was rejected or
+    ///   because the quorum was not reached.
+    /// @dev throws if vote is still ongoing or proposal does not exist
+    function getResult(uint256 proposalIndex) public returns(bool hasPassed) {
+        require(proposalIndex < _proposals.length, "Invalid proposal index");
+        Proposal storage p = _proposals[proposalIndex];
+        require(p.endTime < now, "Voting period is not over");
+
+        uint256 minParticipation = TOKEN.totalSupplyAt(p.snapshotId) * 40 / 100;
```
djudjuu: **NOTE** The current solution assumes the total supply to be not maxed out, otherwise the first product will overflow. Alternatively, if totalSupply * Quorum overflows, one could be okay with small errors at the boundary and do totalSupply / 100 * 40 Alternatively, there would be Library like SafeMath.sol deployed somewhere and we would just use that OR we inherit from Math.sol

test/Voting.js
```diff
@@ -0,0 +1,22 @@
+import { testVotingWithSnapshots } from "./helpers/votingTestCases";
+import { ZERO_ADDRESS } from "./helpers/constants";
+
+// this is just a convenience testfile so I dont have always have to run the entire SnapshotTokentestsuite
```
djudjuu: NOTE: this comment is a reminder to delete this file before merging the PR (if that would be a pr to be merged). The file only serves to accelerate the workflow.

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+    ///   because the quorum was not reached.
+    /// @dev throws if vote is still ongoing or proposal does not exist. Also deletes the proposalData to free up gas.
+    //     Therefore subsequent attempts to call this function will throw too
+    function getResult(uint256 proposalId) public {
+        require(proposalId < _nextProposalId, "Invalid proposal ID");
+        Proposal storage p = _proposals[proposalId];
+        require(p.owner != address(0), "Proposal was already deleted");
+        require(p.endTime < now, "Voting period is not over");
+
+        uint256 minParticipation = TOKEN.totalSupplyAt(p.snapshotId) * 40 / 100;
```
djudjuu: see assumptions/things to discuss iii)

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+        bool passedQuorum = p.inFavor + p.against > minParticipation;
+        bool hasPassed = passedQuorum && p.inFavor > p.against;
+        _hasActiveProposal[_proposals[proposalId].owner] = false;
+
+        emit LogProposalResult(proposalId, hasPassed, p.inFavor, p.against);
+
+        delete _proposals[proposalId];
```
djudjuu: According to CodeStyle.md writing the log should come after writing to storage. My personal preference is to to deletions to free up gas after everything else. I could change that though

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+        address voter,
+        bytes32 r,
+        bytes32 s,
+        uint8 v
+    )
+        public
+    {
+        // check that message signature matches the voter address
+        // solium-disable indentation
+        require(ecrecover(
```
rudolfix: 👍 for following "old" EIP but not getting into https://github.com/alexvandesande/EIPs/blob/ee2347027e94b93708939f2e448447d030ca2d76/EIPS/eip-1077.md and sign typed data as for example Gnosis Safe is doing. 
the objective here was to minimize gas usage to make voting affordable and and some lean format is the way to do it

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+            keccak256(abi.encodePacked(
+                "\x19Ethereum Signed Message:\n32",
```
rudolfix: we could probably optimize this part by pre-encoding `"\x19Ethereum Signed Message:\n32", keccak256(abi.encodePacked(byte(0)` but it would (probably) involve some inline assembly trickery. 
do you know if this `keccak256(abi.encodePacked(byte(0)` will be evaluated at compile time?

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+                keccak256(abi.encodePacked(byte(0), this, proposalId, voteInFavor)))),
```
rudolfix: it's a good practice to `case` this to `address`, AFAIK solidity 5.x would not allow this at all (not 100% however)

contracts/SnapshotToken/Extensions/SimpleVote.sol
rudolfix: your explanations are sufficient. I typically put overflow assumptions in the comments whenever I ignore some impossible edge cases ;>

here to be bullet proof, you could use `Math.sol` (that you also mention) and check out `proportion` method which was done exactly for that purpose and has full test coverage...

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+    /// @dev throws if vote is still ongoing or proposal does not exist. Also deletes the proposalData to free up gas.
+    //     Therefore subsequent attempts to call this function will throw too
+    function getResult(uint256 proposalId) public {
+        require(proposalId < _nextProposalId, "Invalid proposal ID");
+        Proposal storage p = _proposals[proposalId];
+        require(p.owner != address(0), "Proposal was already deleted");
+        require(p.endTime < now, "Voting period is not over");
+
+        uint256 minParticipation = TOKEN.totalSupplyAt(p.snapshotId) * 40 / 100;
+        bool passedQuorum = p.inFavor + p.against > minParticipation;
```
rudolfix: non inclusive range is OK here - thanks for explanations!
practically in most governance systems (at least for companies) quorums are inclusive

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+        bool hasPassed = passedQuorum && p.inFavor > p.against;
```
rudolfix: cool! works well for > 50% majority

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+        require(!p.hasVoted[voter], "Address has already voted");
+        uint256 weight = TOKEN.balanceOfAt(voter, p.snapshotId);
+        require(weight > 0, "Token balance at proposal time is zero");
+
+        if (voteInFavor) {
+            p.inFavor += weight;
+        } else {
+            p.against += weight;
+        }
+        p.hasVoted[voter] = true;
```
rudolfix: :+1: as this protect against re-using the same message by the relayer or any malicious actor as relayer messages has no nonce.

architectural note:
that is good solution specific to this contract as it minimizes gas usage for relayed messages. HOWEVER if we push such thing to production we would probably provide some mixin class that you would derive from in the contracts that use meta-transaction. such mixin would provide nonce storage and force some basic message structure that has a nonce.

test/Voting.js
rudolfix: oh this is actually good! I was probably not clear in my task description or I just wanted to minimize your work on tests.
actually you could remove those tests from `SnapshotToken.js` and just leave this file (put it into `extras` folder and rename to `SimpleVoting.js`)

test/helpers/votingTestCases.js
```diff
@@ -0,0 +1,188 @@
+import { expect } from "chai";
+import { increaseTime } from "../helpers/evmCommands";
+import { createSignedVote } from "../helpers/relayedVoteSigning";
+import { prettyPrintGasCost } from "../helpers/gasUtils";
+
+export function testVotingWithSnapshots(getToken, getVotingContract, owner, owner2, relayer) {
```
rudolfix: yeah big 👍 for test coverage. actually we use form of test suites that can be used against any contract just for the standard interfaces (like ERC20. Snapshot, TokenController etc.) but it's pretty cool that you did it this way, following this pattern.

test/helpers/votingTestCases.js
```diff
@@ -0,0 +1,188 @@
...
+
+  beforeEach(() => {
+    votingContract = getVotingContract();
+    token = getToken();
+  });
+
+  describe("Opening Proposals:", () => {
+    it("should allow to open a proposal", async () => {
+      const addProposalLog = await votingContract.addProposal();
+      await expect(addProposalLog.logs.find(log => log.event === "LogNewProposal")).to.not.be.empty;
```
rudolfix: actually we have a set of nice helper functions for testing logs. and we test every field in the log of course.

test/helpers/votingTestCases.js
```diff
@@ -0,0 +1,188 @@
...
+    });
+
+    it("should reject opening of more than one proposal per address", async () => {
+      const addProposalLog = await votingContract.addProposal();
+      await expect(addProposalLog.logs.find(log => log.event === "LogNewProposal")).to.not.be.empty;
+      await expect(votingContract.addProposal()).to.be.rejectedWith(
```
rudolfix: 👍 for checking revert code

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+                v[i]
+            );
+        }
+    }
+
+    /// @return whether the vote on a proposal has passed, if false this could be because it was rejected or
+    ///   because the quorum was not reached.
+    /// @dev throws if vote is still ongoing or proposal does not exist. Also deletes the proposalData to free up gas.
+    //     Therefore subsequent attempts to call this function will throw too
+    function getResult(uint256 proposalId) public {
```
rudolfix: practical note:
a `constant` version of this function would be pretty useful to dapp developers (if we want to show results during the voting).

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+        uint256 weight = TOKEN.balanceOfAt(voter, p.snapshotId);
+        require(weight > 0, "Token balance at proposal time is zero");
+
+        if (voteInFavor) {
+            p.inFavor += weight;
+        } else {
+            p.against += weight;
+        }
+        p.hasVoted[voter] = true;
+    }
```
rudolfix: practical notes
you should log event in here. that event could be for example picked up by our backend to send you confirmation e-mail that you voted, we also have a relational mapper microservice that would convert this event and write to relational database which you can query for aggregates and statistics. also note that such event is very useful for testing as you do not expose any storage

you also do not store who voted how and that's good (gas)! with the log event voting actors can easily check how they voted (also good for the dapp developers)

test/helpers/votingTestCases.js
```diff
@@ -0,0 +1,188 @@
...
+  let token;
+  let votingContract;
+
+  beforeEach(() => {
+    votingContract = getVotingContract();
+    token = getToken();
+  });
+
+  describe("Opening Proposals:", () => {
+    it("should allow to open a proposal", async () => {
```
rudolfix: practical note:
you should also test public properties of the contract after it has been deployed. is `TOKEN` set correctly? is voting period as expected?

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+    function incrementVote(uint256 proposalId, bool voteInFavor, address voter) internal {
+        require(proposalId < _nextProposalId, "Invalid proposal ID");
+        Proposal storage p = _proposals[proposalId];
+        require(p.endTime > now, "Voting period is over");
+        require(!p.hasVoted[voter], "Address has already voted");
+        uint256 weight = TOKEN.balanceOfAt(voter, p.snapshotId);
+        require(weight > 0, "Token balance at proposal time is zero");
+
+        if (voteInFavor) {
+            p.inFavor += weight;
```
rudolfix: just one small note (I discovered this recently). you read and write storage here. in that case solidity optimizer will be unable to reduce number of writes to storage. the patterns is to read storage and then write it in one block.

here it does not matter - we use full words to store vote weights

contracts/SnapshotToken/Extensions/SimpleVote.sol
rudolfix: ok if you provide `voter` in parameter you should check if the erecover gives you same address

now actually relayer can use valid message to vote on someone else as `voter` is not a part of the message

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+        uint256 snapshotId;
+        mapping (address => bool) hasVoted;
+    }
+
+    /////////////////////////
+    // Immutable state
+    ////////////////////////
+    ISnapshotableToken public TOKEN;
+    // % of tokens to participate in a voting to make it valid
+    uint256 public QUORUM = 40;
```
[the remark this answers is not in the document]
djudjuu: oh yeah, I used 40 in the code... now it is using the variable

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+    uint256 public VOTING_PERIOD;
+
+    /////////////////////////
+    // Mutable state
+    ////////////////////////
+    mapping (uint256 => Proposal) private _proposals;
```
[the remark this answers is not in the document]
djudjuu: I thought one way of learning about hte proposal (events) is enough. The second just wastes gas. so yeah, this was by design.... if we acually planned to build a vote-ui this could be reconsidered, but so far the design philosophy is to save gas

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+
+    /// @param token which balances are used for voting
+    /// @param votingPeriodInDays amount of days that each vote will take
+    constructor(
+        ISnapshotableToken token,
+        uint256 votingPeriodInDays
+    )
+        public
+    {
+        VOTING_PERIOD = votingPeriodInDays * 1 days;
```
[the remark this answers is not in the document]
djudjuu: done

contracts/SnapshotToken/Extensions/SimpleVote.sol
djudjuu: yeah, here my design choice, was to not make intermediate results easily accessible and to save gas

contracts/SnapshotToken/Extensions/SimpleVote.sol
```diff
@@ -0,0 +1,191 @@
...
+        incrementVote(proposalId, voteInFavor, msg.sender);
+    }
+
+    /// @dev same as vote, only for a relayed vote. Will throw if provided signature (v,r,s) does not match
+    ///  the address of the voter
+    /// @param voter address whose token balance should be used as voting weight
+    function relayedVote(
+        uint256 proposalId,
+        bool voteInFavor,
+        address voter,
```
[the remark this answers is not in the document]
djudjuu: true, but we would lose helpful error message, in case the signature was not constructed the right way....

PR #316, opened 2019-11-07
raise block gas limit to 8 million

djudjuu:
while the current limit is at 10 million (https://ethgasstation.info/blog/ethereum-block-size/)
I chose to only raise to 8, so that we will be conscious of any changes to increase our contract size

## NOTE: For this to be effective the gas limit has to be raised on our parity-clone as well, see PR
https://github.com/Neufund/parity-instant-seal-byzantium-enabled/pull/25

## Motivation:

adding an extra convenience function to ``` MockEtoCommitment.sol``` was not possible as it would go above the block gas limit

### DIFF
--- a/truffle.js
+++ b/truffle.js
@@ -63,7 +63,7 @@ module.exports = {
   networks: {
     localhost: {
       network_id: "*",
-      gas: 6800000,
+      gas: 8000000,
       gasPrice: 21000000000,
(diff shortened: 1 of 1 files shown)


djudjuu:
I suspect the test is failing because it uses the parity-node in the background, which has not raised the GL yet. So then the parity would have to be merged first.

PR #317, opened 2019-11-07
shiftTime Helper (+raise gas limit)

djudjuu:
a simple helper + test to shift time on an ETO to right before the next state transition.

NOTE: 
I think, for this to be effective the gas limit has to be raised on our parity-clone as well, see PR
Neufund/parity-instant-seal-byzantium-enabled#25

While the current limit is at 10 million (https://ethgasstation.info/blog/ethereum-block-size/)
I chose to only raise to 8, so that we will be conscious of any changes to increase our contract size


djudjuu:
please do not merge yet, I am seeing if i can refactor it to define that variable in one place only and then use it everywhere. 

(but historically, this has been difficult to do in truffle.js, somehow)

contracts/test/ETO/MockETOCommitment.sol
```diff
@@ -43,6 +43,15 @@ contract MockETOCommitment is
         }
     }
 
+    // convenience function for moving all timestampts towards the past
+    // such that the next state transition will occur in delta seconds
+    function _shiftToBeforeNextState(uint8 delta) public {
+        require(delta < 180);
```
rudolfix: so 3 minutes is good for everyone? also for QA?

truffle.js
```diff
@@ -104,7 +108,7 @@ module.exports = {
     },
     forked_live: {
       network_id: 72,
-      gas: 6800000, // close to current mainnet limit
+      gas: gasLimit,
```
rudolfix: please keep old value on `forked` networks - they use different instance where gas floor is `6800000`, alternatively you can define another variable

you can also use same value for mainnet. I know that block limit is not much higher but we should not have any contract or transaction that goes above `forked` (staging) limit and we can safely enforce it.

contracts/test/ETO/MockETOCommitment.sol
```diff
@@ -43,6 +43,15 @@ contract MockETOCommitment is
...
+        uint256 nextTransition = startOfInternal(ETOState(uint(state()) + 1));
+        require(nextTransition != 0 && nextTransition > now + delta);
```
rudolfix: please use revert code like `NF_MOCK_INVALID_DELTA` or similar. it will be easier to debug.

test/ETO/ETOCommitment.js
```diff
@@ -370,6 +370,35 @@ contract("ETOCommitment", ([, admin, company, nominee, ...investors]) => {
       await investAmount(investors[0], Q18.mul(23987.288912), "EUR");
       expect(await etoCommitment.state()).to.be.bignumber.eq(CommitmentState.Public);
     });
+
+    it("should move time to right before state transition", async () => {
```
rudolfix: thx for the test!

contracts/test/ETO/MockETOCommitment.sol
djudjuu: 1800 it is now, approved by @glupiawymowka

contracts/test/ETO/MockETOCommitment.sol
djudjuu: I thought so too, but I never so any revert-errormessages on any of the rquirements so I didn't. That could be something to do the next we do any changes to the contracts..

PR #321, opened 2019-11-27
J!/test receiving

### LINKED ISSUE #320: implement mock that accepts ETH and accepts/rejects ERC223 transfers
**Why we do it**
Due to Istanbul fork we cannot use `SimpleExchange` to test ETH transfer contract to contract. (we had event there that now crosses 2300 gas stipend of fallback function)
We need to reintroduce those e2e tests

**What we do**
We need a contract that will be deployed on dev (loclahost) env against which we can test ETH transfer. On top of that let's add ERC223 callback for testing ERC223 transfer for which we do not have coverage at all

**Acceptance criteria**
* [ ] implements payable callback function
* [ ] function above can be called via another contract (test!)
* [ ] implements IERC223Callback
* [ ] implements IERC223LegacyCallback (to test NEU which requires old callback ;/)
* [ ] provides a mock method to accepts/reject erc223 transfers transfers in the callback function

**Deployment**
* [ ] make it a part of deployment pipeline so it gets deployed only on dev.

djudjuu:
closes #320

### DIFF
--- a/contracts/test/TestReceiver.sol
+++ b/contracts/test/TestReceiver.sol
@@ -9,7 +9,7 @@ import "../Standards/IERC223LegacyCallback.sol";
  - ERC223 tokens
  - tokens with legacy ERC223 tokenFallback (e.g. Neumark.sol)
 */
-contract TestReceiving is IERC223Callback, IERC223LegacyCallback {
+contract TestReceiver is IERC223Callback, IERC223LegacyCallback {
 
     ////////////////////////
     // Mutable state
@@ -34,12 +34,6 @@ contract TestReceiving is IERC223Callback, IERC223LegacyCallback {
     ////////////////////////
     function setERC223Acceptance(bool acceptERC223) public {
         _acceptERC223 = acceptERC223;
-    }
-
-
-    // convenience function to check the balance of this contract
-    function returnBalance() public view returns(uint) {
-        return address(this).balance;
     }
 
     // fallback function to receive ether
--- a/contracts/test/TestReceiving.sol
+++ b/contracts/test/TestReceiving.sol
@@ -1,0 +1,80 @@
+pragma solidity 0.4.26;
+
+
+import "../Standards/IERC223Callback.sol";
+import "../Standards/IERC223LegacyCallback.sol";
+
+/** contract to be used in frontend testing for receiving of
+ - ether
+ - ERC223 tokens
+ - tokens with legacy ERC223 tokenFallback (e.g. Neumark.sol)
+*/
+contract TestReceiving is IERC223Callback, IERC223LegacyCallback {
+
+    ////////////////////////
+    // Mutable state
+    ////////////////////////
+    address private _from;
+    uint256 private _amount;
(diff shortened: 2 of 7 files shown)


djudjuu:
here are my notes taken during this task (markdown-export of emacs-org-mode, feel free to ask for the org-mode version)

==========


# Table of Contents

1.  [build testing contract for FE](#orga996934)
    1.  [?! re: acceptance criteria:](#orgdbefbf6)
        1.  [why can be called by another contract? > check for unviolated gas limits](#orgd98991d)
    2.  [what to build](#org0ba4b95)
    3.  [steps](#org9b314e4)
        1.  [setup tests structure](#org15326f2)
        2.  [unit tests](#orgb8a7762)
        3.  [integration](#org9808d2f)
    4.  [test pipeline](#org52fed61)


<a id="orga996934"></a>

# DONE build testing contract for FE

sending a tx should either fail or not
research: what are the keywords that will enable/disable sending of tokens


<a id="orgdbefbf6"></a>

## ?! re: acceptance criteria:


<a id="orgd98991d"></a>

### why can be called by another contract? > check for unviolated gas limits

other contracts might send ether to it via .send()/transfer which forwards
only 2300gas, which could be potentially too little if the fallback does
expensive operations (note: the previous simpleExchange was doing some
read-calls, probably it went above that limit becasue we were checking it's
own balance, but wihtout the new SELDBALANCE opcode, as we force it being
compiled by an older version of the compiler)

> at some point we might want to consider upgrading to a later solidity-version!


<a id="org0ba4b95"></a>

## TODO what to build

-   contract DONE
-   DONE deploy it: 
    -   in which file?
    -   in "101 apply fixtures" simple import: "const TestReceiving = artifacts.require("TestReceiving")"
    -   ?! how to run the deployer in there? TODO deploy() doe snot work
        -   "network state unknwon" / contract.detectNetwork is not a function > SPELLING mistake
    -   ?? how to assure that it always ends up at the same address > gets exported to the fixtures?
-   test itDONE :
    -   where to place the test?
    -   create a mock folder for the tests
    -   beforeEach(): deploy Receiver


<a id="org9b314e4"></a>

## steps


<a id="org15326f2"></a>

### DONE setup tests structure


<a id="orgb8a7762"></a>

### DONE unit tests

-   do unit tests (jsut call the fallbacks directly)
-   !! use transfer() not send() to make ether reception revert!


<a id="org9808d2f"></a>

### DONE integration

-   do integration test where I deploy the neutoken & ethertoken in its seperate scope (describe)
    -   arbiter & all other addresses can just be somethting irrelevant


<a id="org52fed61"></a>

## test pipeline

 in contracts:
$ yarn truffle test &#x2013;network inprocess<sub>test</sub> test/Mock.js test/setup.js

hint: 
avoid recompilation of every all contracts before re-running tests by doing:
run npx truffle migrate &#x2013;reset &#x2013;network inprocess<sub>test</sub>

    in FE:
$ yarn start remote, 
$ yarn test:e2e:dev -> somehow this runs all the tests and fails at a differnt one

PR #323, opened 2020-01-15
refunds eto with transfer

rudolfix:
we switch to regular ERC20 transfer in refund operation of ETOCommitment. ERC223 is not supported anywhere anymore it seems and using this will cause problem from smart contract based wallets like Gnosis Safe which do not implement required callbacks in their proxy contracts

### DIFF
--- a/contracts/ETO/ETOCommitment.sol
+++ b/contracts/ETO/ETOCommitment.sol
@@ -985,6 +985,7 @@ contract ETOCommitment is
         if (ticket.equivEurUlps == 0) {
             return;
         }
+        // modify state before sending
         ticket.claimOrRefundSettled = true;
 
         if (ticket.rewardNmkUlps > 0) {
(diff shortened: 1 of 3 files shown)


djudjuu:
hmm. it looks good to me, although I got to admit that I am still a bit confused when I have to look at the ETOCommitment-contract. this time I was taking a while to understand what LockedAccount is doing. But assuming that this was working correctly before, the change you made will call the ERC20-transfer instead of the ERC223-transfer. 

I am not sure if there are tests where the investor(-wallet) is a contract and does not have a tokenFallback and we expect it to fail.... but that would be a change to keep the testsuite up2date....which I guess is okay if we dont, unless we want to merge this to production...

rudolfix:
writing a test where we do whole ETO flow for a wallet which is a smart contract is a very good idea. deploying full gnosis safe or similar is probably too much but we could write a simple forwarding contract to see how system behaves.

there will be ticket

PR #324, opened 2020-01-15
[EXTRAS] implements extra equity token controllers

rudolfix:
extra code with with two token controllers
1. granular transfer controller where you can do forced transfers and freeze/unfreeze accounts
2. reg d controller where token lockup period is implemented

extra code is not intended for production deployment, finally we want modularized controllers and some higher level interpreter for executing token holder and issuer rights


contracts/Company/Extras/GranularTransferController.sol
```diff
@@ -0,0 +1,187 @@
...
+            return _forcedTransfers[owner].amount;
+        }
+    }
+
+    //
+    // Overrides IContract
+    //
+
+    function contractId() public pure returns (bytes32 id, uint256 version) {
+        return (0xf7e00d1a4168be33cbf27d32a37a5bc694b3a839684a8c2bef236e3594345d70, 0xFF);
```
djudjuu: what is this address? I think a comment would be helpful and make it maintanable

contracts/Company/Extras/GranularTransferController.sol
```diff
@@ -0,0 +1,187 @@
...
+        _frozenAddresses[owner] = true;
+        emit LogTokenHolderAccountFrozen(owner);
+
+    }
+
+    function unfreezeHolder(address owner)
+        public
+        onlyCompany
+    {
+        _frozenAddresses[owner] = true;
```
djudjuu: this should proably be ```false```

contracts/Company/Extras/GranularTransferController.sol
```diff
@@ -0,0 +1,187 @@
...
+        // allow forced transfer only by the token controller itself
+        if ((s == GovState.Funded || s == GovState.Closing) && broker == address(this)) {
+            ForcedTransfer storage t = _forcedTransfers[from];
+            // check if forced transfer matches the actual transfer
+            if (t.amount > 0) {
+                // we allow because we checked the following
+                //  (1) the broker is the company
+                //  (2) there's forced transfer defined for from address
+                //  (3) there's a match for to and amount
+                allow = t.amount == amount && t.to == to;
```
djudjuu: shouldnt this be:

```allow = allow && (t.amount == amount && t.to == to);```

?

I dont prefectely understand the conditions including GovState but even if they only pass if allow is already true it looks very implicit

File: truffle.js [hunk 4]
```diff
@@ -166,22 +167,22 @@ module.exports = {
...
+        "https://ndfull.neufund.org/",
         // "44'/60'/0'/0",
         // "44'/60'/105'/7", // identity management (A)
         // "44'/60'/105'/3", // reclaimer
-        // "44'/60'/105'/0", // legal rep (M)
+        "44'/60'/105'/0", // legal rep (M)
         // "44'/60'/105'/2", // eurt legal manager (M)
         // "44'/60'/105'/11", //DEPLOYER (admin)
-        "44'/60'/106'/16",
+        // "44'/60'/106'/16",
```
djudjuu: I dont understand why this needs to be done, so I cant really review this....

test/ETO/Extras/ExtraTokenControllers.js
```diff
@@ -0,0 +1,352 @@
...
+            "1",
+          );
+          expect(canTransfer).to.be.false;
+          // todo: test forced transfer interaction, but that's for the real code
+          // now unfreeze
+          await equityTokenController.unfreezeHolder(investor1, { from: company });
+          // receive
+          await equityToken.transfer(investor2, 1, { from: investor1 });
+          // send
+          await equityToken.transfer(investor1, 1, { from: investor2 });
```
djudjuu: I would prefer to have actual assert-statements here, not just executed transfers: e.g.
```
          await equityToken.transfer(investor2, 1, { from: investor1 });
          expect(await equityToken.balanceOf(investor2)).to.be.bignumber.eq(2);
          // send
          await equityToken.transfer(investor1, 1, { from: investor2 });
          expect(await equityToken.balanceOf(investor2)).to.be.bignumber.eq(1);
```

test/ETO/Extras/ExtraTokenControllers.js
```diff
@@ -0,0 +1,352 @@
...
+          canTransfer = await equityTokenController.onTransfer(
+            investor1,
+            investor1,
+            investor2,
+            "1",
+          );
+          expect(canTransfer).to.be.false;
+        });
+      });
+    });
```
djudjuu: for production i'd want some more tests specific on 
```if("should freeze accounts such that they can no longer transfer")```
and 
```if("should unfreeze frozen accounts such that they transfer again")```

File: test/ETO/Extras/ExtraTokenControllers.js [hunk 7]
```diff
@@ -0,0 +1,352 @@
...
+        await identityRegistry.setClaims(investor1, toBytes32("0x0"), toBytes32("0x31"), {
+          from: admin,
+        });
+        // simulate ETO
+        await deployController(RegDTransferController, { ENABLE_TRANSFERS_ON_SUCCESS: true });
+        await deployETO();
+        await makeInvestment(totalShares, inv1DistAmount);
+        // investor 2 is a regular investor with transfer rights
+        await testCommitment._distributeTokens(investor2, "1");
+        // should not be able to send
```
djudjuu: maybe add ```// investor 1 should not be able to send```

this is really tiny detail, but like this the comment is more confusing than helping, because it reads as if the two comments are connected: 

"investor 2 is a regular investor AND should not be able to send

we can leave it though.... i just want to mention it

contracts/Company/Extras/GranularTransferController.sol
rudolfix: but you are right - it does not make sense to obtain `allow` from base class if the whole condition does not need it! it should be done when this is really needed.

saves gas. will fix it.

PR #330, opened 2020-03-02
parametrized equity token decimals

rudolfix:
This PR adds ability to send decimals for equity tokens. Currently we assumed that all equity tokens are indivisible (0 decimals), however DeFi applications and other dapps often assume 18 decimals to be a standard.
On top of that we solve a few rounding issues where for given amount of EUR non exact number of tokens could be purchased - and we were rounding down.

Contracts and tests are updated to allow any number of decimals. Several additional rounding issues were found and fixed in tests.

See `ETOTerms::calculateTokenAmount` and `ETOTerms::calculateEurUlpsAmount` for new formulas to calculate token amount and the reverse. `EQUITY_TOKEN_POWER` is `10**decimals` which was `1` for indivisible tokens, giving the old formula. All tests were updated accordingly.

See commit lists for the exact scope

### DIFF
--- a/contracts/ETO/ETOCommitment.sol
+++ b/contracts/ETO/ETOCommitment.sol
@@ -48,11 +48,12 @@ contract ETOCommitment is
     struct InvestmentTicket {
         // euro equivalent of both currencies.
         //  for ether equivalent is generated per ETH/EUR spot price provided by ITokenExchangeRateOracle
+        // 128bit values 2**128 / 10**18 which is 340 quintillion 282 quadrillion 366 trillion 920 billion 938 million 463 thousand 463 nEUR max
         uint96 equivEurUlps;
         // NEU reward issued
         uint96 rewardNmkUlps;
         // Equity Tokens issued, no precision
-        uint96 equityTokenInt;
+        uint96 equityTokenAmount;
         // total Ether invested
         uint96 amountEth;
         // total Euro invested
@@ -100,6 +101,8 @@ contract ETOCommitment is
     uint256 private MAX_INVESTMENT_AMOUNT_EUR_ULPS;
     // min ticket size, taken from eto terms
     uint256 private MIN_TICKET_EUR_ULPS;
+    // equity token power (10^precision)
+    uint256 private EQUITY_TOKEN_POWER;
 
     // company representative address
     address private COMPANY_LEGAL_REPRESENTATIVE;
@@ -122,35 +125,35 @@ contract ETOCommitment is
     // investment tickets
     mapping (address => InvestmentTicket) private _tickets;
 
-    // data below start at 32 bytes boundary and pack into 32 bytes word
+    // data below start at 32 bytes boundary and pack into two 32 bytes words
     // total investment in euro equivalent (ETH converted on spot prices)
-    uint112 private _totalEquivEurUlps;
+    uint128 private _totalEquivEurUlps;
     // total equity tokens acquired
-    uint56 private _totalTokensInt;
+    uint128 private _totalTokenAmount;
     // total equity tokens acquired in fixed slots
-    uint56 private _totalFixedSlotsTokensInt;
+    uint128 private _totalFixedSlotsTokenAmount;
     // total investors that participated
-    uint32 private _totalInvestors;
+    uint128 private _totalInvestors;
 
     // nominee investment agreement url confirmation hash
     bytes32 private _nomineeSignedInvestmentAgreementUrlHash;
 
     // additonal contribution / investment amount eth
     // it holds investment eth amount until end of public phase, then additional contribution
-    uint112 private _additionalContributionEth;
+    uint128 private _additionalContributionEth;
     // additonal contribution / investment amount eur
     // it holds investment eur amount until end of public phase, then additional contribution
-    uint112 private _additionalContributionEurUlps;
+    uint128 private _additionalContributionEurUlps;
 
     // successful ETO bookeeping
-    // amount of new shares generated never exceeds number of tokens (uint56)
-    uint64 private _newShares;
-    // how many equity tokens goes to platform portfolio as a fee
-    uint64 private _tokenParticipationFeeInt;
+    // amount of new shares generated never exceeds number of tokens (uint128)
+    uint128 private _newShares;
     // platform fee in eth
-    uint112 private _platformFeeEth;
+    uint128 private _platformFeeEth;
     // platform fee in eur
-    uint112 private _platformFeeEurUlps;
+    uint128 private _platformFeeEurUlps;
+    // how many equity tokens goes to platform portfolio as a fee
+    uint128 private _tokenParticipationFee;
 
     // signed investment agreement url
     string private _signedInvestmentAgreementUrl;
@@ -251,6 +254,7 @@ contract ETOCommitment is
         MAX_AVAILABLE_TOKENS_IN_WHITELIST = etoTerms.MAX_AVAILABLE_TOKENS_IN_WHITELIST();
         MIN_NUMBER_OF_TOKENS = etoTerms.TOKEN_TERMS().MIN_NUMBER_OF_TOKENS();
         MIN_TICKET_TOKENS = etoTerms.calculateTokenAmount(0, MIN_TICKET_EUR_ULPS);
+        EQUITY_TOKEN_POWER = etoTerms.EQUITY_TOKEN_POWER();
 
         MAX_INVESTMENT_AMOUNT_EUR_ULPS = ETO_TERMS_CONSTRAINTS.MAX_INVESTMENT_AMOUNT_EUR_ULPS();
         // set it to max(unit256) to reduce number of operations later
@@ -347,8 +351,8 @@ contract ETOCommitment is
         acceptAgreementInternal(investor);
         // check eligibilty, terms and issue NEU + ET
         (,,,,
-            uint256 equityTokenInt,
-            uint256 fixedSlotEquityTokenInt,
+            uint256 equityTokenAmount,
+            uint256 fixedSlotEquityTokenAmount,
             uint256 investorNmk) = issueTokens(investor, wallet, equivEurUlps);
 
         // update investor ticket
@@ -357,14 +361,14 @@ contract ETOCommitment is
             wallet,
             equivEurUlps,
             investorNmk,
-            equityTokenInt,
+            equityTokenAmount,
             amount
         );
 
         // update investment state
         updateTotalInvestment(
-            equityTokenInt,
-            fixedSlotEquityTokenInt,
+            equityTokenAmount,
+            fixedSlotEquityTokenAmount,
             equivEurUlps,
             amount,
             firstTimeInvestment
@@ -377,7 +381,7 @@ contract ETOCommitment is
             msg.sender,
             amount,
             equivEurUlps,
-            equityTokenInt,
+            equityTokenAmount,
             EQUITY_TOKEN,
             investorNmk
         );
@@ -435,14 +439,14 @@ contract ETOCommitment is
         returns (
             uint256 newShares, uint256 capitalIncreaseUlps,
             uint256 additionalContributionEth, uint256 additionalContributionEurUlps,
-            uint256 tokenParticipationFeeInt, uint256 platformFeeEth, uint256 platformFeeEurUlps,
+            uint256 tokenParticipationFee, uint256 platformFeeEth, uint256 platformFeeEurUlps,
             uint256 sharePriceEurUlps
         )
     {
         return (
             _newShares, _newShares * ETO_TERMS.TOKEN_TERMS().SHARE_NOMINAL_VALUE_ULPS(),
             _additionalContributionEth, _additionalContributionEurUlps,
-            _tokenParticipationFeeInt, _platformFeeEth, _platformFeeEurUlps,
+            _tokenParticipationFee, _platformFeeEth, _platformFeeEurUlps,
             _newShares == 0 ? 0 : divRound(_totalEquivEurUlps, _newShares)
         );
     }
@@ -482,11 +486,11 @@ contract ETOCommitment is
         constant
         returns (
             uint256 totalEquivEurUlps,
-            uint256 totalTokensInt,
+            uint256 totalTokenAmount,
             uint256 totalInvestors
             )
     {
-        return (_totalEquivEurUlps, _totalTokensInt, _totalInvestors);
+        return (_totalEquivEurUlps, _totalTokenAmount, _totalInvestors);
     }
 
     function calculateContribution(address investor, bool fromIcbmWallet, uint256 newInvestorContributionEurUlps)
@@ -499,21 +503,20 @@ contract ETOCommitment is
             bool isEligible,
             uint256 minTicketEurUlps,
             uint256 maxTicketEurUlps,
-            uint256 equityTokenInt,
+            uint256 equityTokenAmount,
             uint256 neuRewardUlps,
-            bool maxCapExceeded,
-            uint256 spentAmountEurUlps
+            bool maxCapExceeded
             )
     {
         InvestmentTicket storage ticket = _tickets[investor];
-        uint256 fixedSlotsEquityTokenInt;
+        uint256 fixedSlotsEquityTokenAmount;
         (
             isWhitelisted,
             isEligible,
             minTicketEurUlps,
             maxTicketEurUlps,
-            equityTokenInt,
-            fixedSlotsEquityTokenInt
+            equityTokenAmount,
+            fixedSlotsEquityTokenAmount
         ) = ETO_TERMS.calculateContribution(
             investor,
             _totalEquivEurUlps,
@@ -527,9 +530,12 @@ contract ETOCommitment is
             (,neuRewardUlps) = calculateNeumarkDistribution(NEUMARK.incremental(newInvestorContributionEurUlps));
         }
         // crossing max cap can always happen
-        maxCapExceeded = isCapExceeded(state() == ETOState.Whitelist, equityTokenInt, fixedSlotsEquityTokenInt, newInvestorContributionEurUlps);
-        // todo: take it from ETO_TERMS.calculateContribution
-        spentAmountEurUlps = newInvestorContributionEurUlps;
+        maxCapExceeded = isCapExceeded(
+            state() == ETOState.Whitelist,
+            equityTokenAmount,
+            fixedSlotsEquityTokenAmount,
+            newInvestorContributionEurUlps
+        );
     }
 
     function investorTicket(address investor)
@@ -538,7 +544,7 @@ contract ETOCommitment is
         returns (
             uint256 equivEurUlps,
             uint256 rewardNmkUlps,
-            uint256 equityTokenInt,
+            uint256 equityTokenAmount,
             uint256 sharesFrac,
             uint256 tokenPrice,
             uint256 neuRate,
@@ -552,9 +558,9 @@ contract ETOCommitment is
         // here we assume that equity token precisions is 0
         equivEurUlps = ticket.equivEurUlps;
         rewardNmkUlps = ticket.rewardNmkUlps;
-        equityTokenInt = ticket.equityTokenInt;
-        sharesFrac = ETO_TERMS.equityTokensToShares(ticket.equityTokenInt);
-        tokenPrice = equityTokenInt > 0 ? equivEurUlps / equityTokenInt : 0;
+        equityTokenAmount = ticket.equityTokenAmount;
+        sharesFrac = ETO_TERMS.equityTokensToSharesFrac(ticket.equityTokenAmount);
+        tokenPrice = equityTokenAmount > 0 ? ETO_TERMS.calculateTokenEurPrice(equivEurUlps, equityTokenAmount) : 0;
         neuRate = rewardNmkUlps > 0 ? proportion(equivEurUlps, 10**18, rewardNmkUlps) : 0;
         amountEth = ticket.amountEth;
         amountEurUlps = ticket.amountEurUlps;
@@ -623,7 +629,7 @@ contract ETOCommitment is
     {
         // force refund if floor criteria are not met
         // todo: allow for super edge case when MIN_NUMBER_OF_TOKENS is very close to MAX_AVAILABLE_TOKENS and we are within minimum ticket
-        if (newState == ETOState.Signing && _totalTokensInt < MIN_NUMBER_OF_TOKENS) {
+        if (newState == ETOState.Signing && _totalTokenAmount < MIN_NUMBER_OF_TOKENS) {
             return ETOState.Refund;
         }
         // go to refund if attempt to go to Claim without nominee agreement confirmation
@@ -690,44 +696,46 @@ contract ETOCommitment is
         // so if we sold MAX_AVAILABLE_TOKENS that were computed via calculateAmountWithoutFee from MAXIMUM_NUMBER_OF_TOKENS
         // the reverse operation, that is MAX_AVAILABLE_TOKENS + calculatePlatformTokenFee will not always be MAXIMUM_NUMBER_OF_TOKENS
         // while it's probably possible to detect it we put this check here
-        uint256 tokenParticipationFeeInt;
-        if (_totalTokensInt == MAX_AVAILABLE_TOKENS) {
-            uint256 maxTokens = tokenTerms.MAX_NUMBER_OF_TOKENS();
+        uint256 tokenParticipationFee;
+        uint256 maxTokens = tokenTerms.MAX_NUMBER_OF_TOKENS();
+        if (_totalTokenAmount == MAX_AVAILABLE_TOKENS) {
             // rest up until MAX NUMBER OF TOKENS is our fee
             // we also assume that MAX_NUMBER_OF_TOKENS amount to full shares, which token terms contract checks
-            tokenParticipationFeeInt = maxTokens - MAX_AVAILABLE_TOKENS;
+            tokenParticipationFee = maxTokens - MAX_AVAILABLE_TOKENS;
         } else {
-            tokenParticipationFeeInt = PLATFORM_TERMS.calculatePlatformTokenFee(_totalTokensInt);
+            tokenParticipationFee = PLATFORM_TERMS.calculatePlatformTokenFee(_totalTokenAmount);
             // we must have integer number of shares
-            uint256 tokensRemainder = (_totalTokensInt + tokenParticipationFeeInt) % tokensPerShare;
+            uint256 tokensRemainder = (_totalTokenAmount + tokenParticipationFee) % tokensPerShare;
             if (tokensRemainder > 0) {
                 // round up to whole share
-                tokenParticipationFeeInt += tokensPerShare - tokensRemainder;
+                tokenParticipationFee += tokensPerShare - tokensRemainder;
             }
         }
-        // will not cross max number of tokens thats in 2**56 range
-        uint256 totalIssuedTokens = _totalTokensInt + tokenParticipationFeeInt;
+        uint256 totalIssuedTokens = _totalTokenAmount + tokenParticipationFee;
         // round number of shares
-        // require(totalIssuedTokens % tokensPerShare == 0, "NF_MUST_ISSUE_WHOLE_SHARES");
+        require(totalIssuedTokens % tokensPerShare == 0, "NF_MUST_ISSUE_WHOLE_SHARES");
         // we could not cross maximum number of tokens
         // require(totalIssuedTokens <= maxTokens, "NF_FEE_CROSSING_CAP");
-        // todo: optimizer will not handle code below, correction needs to separate storage reads and writes
-        _newShares = uint64(totalIssuedTokens / tokensPerShare);
-        // preserve platform token participation fee to be send out on claim transition
-        _tokenParticipationFeeInt = uint64(tokenParticipationFeeInt);
         // compute fees to be sent on payout transition
-        // 112bit values 2**112 / 10**18 which is 5 quadrillion 192 trillion 296 billion 858 million 534 thousand 827 nEUR max
-        // also we capped balances already when taking from tokens
-        _platformFeeEth = uint112(PLATFORM_TERMS.calculatePlatformFee(_additionalContributionEth));
-        _platformFeeEurUlps = uint112(PLATFORM_TERMS.calculatePlatformFee(_additionalContributionEurUlps));
+        // fees are smaller than total tokens and total nEUR/ETH so diregard overflows
+        uint128 newShares = uint128(totalIssuedTokens / tokensPerShare);
+        uint128 platformFeeEth = uint128(PLATFORM_TERMS.calculatePlatformFee(_additionalContributionEth));
+        uint128 platformFeeEurUlps = uint128(PLATFORM_TERMS.calculatePlatformFee(_additionalContributionEurUlps));
+        // write to storage in one go
+        _newShares = newShares;
+        _platformFeeEth = platformFeeEth;
+        _platformFeeEurUlps = platformFeeEurUlps;
+        // preserve platform token participation fee to be send out on claim transition
+        _tokenParticipationFee = uint128(tokenParticipationFee);
+        // todo: optimizer will not handle code below, correction needs to separate storage reads and writes
         // compute additional contributions to be sent on claim transition
         _additionalContributionEth -= _platformFeeEth;
         _additionalContributionEurUlps -= _platformFeeEurUlps;
         // nominee gets nominal share value immediately to be added to cap table
         uint256 capitalIncreaseEurUlps = tokenTerms.SHARE_NOMINAL_VALUE_EUR_ULPS() * _newShares;
         // limit the amount if balance on EURO_TOKEN < capitalIncreaseEurUlps. in that case Nomine must handle it offchain
-        // no overflow as smaller one is uint112
-        uint112 availableCapitalEurUlps = uint112(min(capitalIncreaseEurUlps, _additionalContributionEurUlps));
+        // no overflow as smaller one is uint128
+        uint128 availableCapitalEurUlps = uint128(min(capitalIncreaseEurUlps, _additionalContributionEurUlps));
         assert(EURO_TOKEN.transfer(NOMINEE, availableCapitalEurUlps, ""));
         // decrease additional contribution by value that was sent to nominee
         _additionalContributionEurUlps -= availableCapitalEurUlps;
@@ -754,7 +762,7 @@ contract ETOCommitment is
             assert(EURO_TOKEN.transfer(COMPANY_LEGAL_REPRESENTATIVE, _additionalContributionEurUlps, ""));
         }
         // issue missing tokens
-        EQUITY_TOKEN.issueTokens(_tokenParticipationFeeInt);
+        EQUITY_TOKEN.issueTokens(_tokenParticipationFee);
         emit LogPlatformNeuReward(TOKEN_OFFERING_OPERATOR, rewardNmk, platformNmk);
         emit LogAdditionalContribution(COMPANY_LEGAL_REPRESENTATIVE, ETHER_TOKEN, _additionalContributionEth);
         emit LogAdditionalContribution(COMPANY_LEGAL_REPRESENTATIVE, EURO_TOKEN, _additionalContributionEurUlps);
@@ -766,15 +774,15 @@ contract ETOCommitment is
     {
         // burn all neumark generated in this ETO (will also burn NEU sent via erroneous/malicious transfers)
         uint256 balanceNmk = NEUMARK.balanceOf(this);
-        uint256 balanceTokenInt = EQUITY_TOKEN.balanceOf(this);
+        uint256 balanceToken = EQUITY_TOKEN.balanceOf(this);
         if (balanceNmk > 0) {
             NEUMARK.burn(balanceNmk);
         }
         // destroy all tokens generated in ETO
-        if (balanceTokenInt > 0) {
-            EQUITY_TOKEN.destroyTokens(balanceTokenInt);
+        if (balanceToken > 0) {
+            EQUITY_TOKEN.destroyTokens(balanceToken);
         }
-        emit LogRefundStarted(EQUITY_TOKEN, balanceTokenInt, balanceNmk);
+        emit LogRefundStarted(EQUITY_TOKEN, balanceToken, balanceNmk);
     }
 
     /// called on transition to ETOState.Payout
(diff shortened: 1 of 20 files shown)


contracts/ETO/ETOTokenTerms.sol
```diff
@@ -73,8 +77,7 @@ contract ETOTokenTerms is Math, IContractId {
         require(keccak256(abi.encodePacked(equityTokenName)) != EMPTY_STRING_HASH);
         require(keccak256(abi.encodePacked(equityTokenSymbol)) != EMPTY_STRING_HASH);
         // overflows cannot be possible
-        require(maxNumberOfTokens < 2**56, "NF_TOO_MANY_TOKENS");
-        require(mul(tokenPriceEurUlps, maxNumberOfTokens) < 2**112, "NF_TOO_MUCH_FUNDS_COLLECTED");
+        require(maxNumberOfTokens < 2**128, "NF_TOO_MANY_TOKENS");
```
djudjuu: what is the reasoning behind increasing to 2**128 here? 

if we want to keep the same limit... it would be something around 2**64, i think...

contracts/ETO/ETOTokenTerms.sol
```diff
@@ -85,14 +88,19 @@ contract ETOTokenTerms is Math, IContractId {
         EQUITY_TOKEN_NAME = equityTokenName;
         EQUITY_TOKEN_SYMBOL = equityTokenSymbol;
         EQUITY_TOKENS_PER_SHARE = equityTokensPerShare;
```
djudjuu: we dont mean ULPS here, do we?

like this: i'd read it as EQUITY_TOKENS_PER_SHARE = 10 && PRECISION = 2  => 1 Share = 1 * 10 * 10^2 = 1000 Tokens

contracts/ETO/ETOTokenTerms.sol
```diff
@@ -85,14 +88,19 @@ contract ETOTokenTerms is Math, IContractId {
...
+        EQUITY_TOKENS_PRECISION = equityTokensPrecision;
+        EQUITY_TOKENS_POWER = 10**uint256(EQUITY_TOKENS_PRECISION);
+
+        require(equityTokensPerShare >= 0 && equityTokensPerShare % EQUITY_TOKENS_POWER == 0, "NF_SHARES_NOT_WHOLE_TOKENS");
```
djudjuu: > equityTokensPerShare >= 0

Why is == 0 allowed ? if not, then 

> equityTokensPerShare % EQUITY_TOKENS_POWER == 0

this only makes sense if ```equityTokensPerShare``` is ULPS, which conflicts with my comment above 

why is any value < EQUITY_TOKENS_POWER allowed? -> For a divisible equity-token to make sense, we need the owners of less than 1 EQUITY-TOKEN to still have a right towards the company, ie. their SHARE needs to be > 0. 
For that to be true

contracts/ETO/ETOTerms.sol
```diff
@@ -199,21 +205,21 @@ contract ETOTerms is
     function calculateTokenAmount(uint256 /*totalEurUlps*/, uint256 committedEurUlps)
```
djudjuu: >     // we require that equity token precision is 0
￼this comment should be removed

contracts/ETO/ETOTerms.sol
```diff
@@ -199,21 +205,21 @@ contract ETOTerms is
...
         public
         constant
-        returns (uint256 tokenAmountInt)
+        returns (uint256 tokenAmount)
     {
         // we may disregard totalEurUlps as curve is flat, round down when calculating tokens
-        return committedEurUlps / calculatePriceFraction(10**18 - PUBLIC_DISCOUNT_FRAC);
+        return mul(committedEurUlps, EQUITY_TOKEN_POWER) / FULL_PRICE_FRACTION;
     }
```
djudjuu: checked.

contracts/ETO/ETOTerms.sol
djudjuu: this comment occures in a bunch of places

contracts/ETO/ETOTerms.sol
```diff
@@ -199,21 +205,21 @@ contract ETOTerms is
...
         constant
         returns (uint256 committedEurUlps)
     {
-        // we may disregard totalTokensInt as curve is flat
-        return mul(tokenAmountInt, calculatePriceFraction(10**18 - PUBLIC_DISCOUNT_FRAC));
+        // we may disregard totalTokenAmount as curve is flat
+        return proportion(tokenAmount, FULL_PRICE_FRACTION, EQUITY_TOKEN_POWER);
     }
 
     function calculatePriceFraction(uint256 priceFrac) public constant returns(uint256) {
```
djudjuu: would be nice to have some documentation here. e.g. TODO

contracts/ETO/ETOTerms.sol
```diff
@@ -395,9 +406,12 @@ contract ETOTerms is
                 // discount is fixed so use base token price
                 if (discountedAmount > 0) {
                     // always round down when calculating tokens
-                    fixedSlotEquityTokenInt = discountedAmount / calculatePriceFraction(wlTicket.fullTokenPriceFrac);
-                    // todo: compute effective amount spent without the rounding
-                    // discountAmount = fixedSlotEquityTokenInt *  calculatePriceFraction(wlTicket.fullTokenPriceFrac);
+                    uint256 fixedSlotPrice = calculatePriceFraction(wlTicket.fullTokenPriceFrac);
+                    fixedSlotEquityTokenAmount = mul(discountedAmount, EQUITY_TOKEN_POWER) / fixedSlotPrice;
```
djudjuu: why should this be higher than before?

contracts/ETO/ETOCommitment.sol
```diff
@@ -48,11 +48,12 @@ contract ETOCommitment is
     struct InvestmentTicket {
         // euro equivalent of both currencies.
         //  for ether equivalent is generated per ETH/EUR spot price provided by ITokenExchangeRateOracle
+        // 128bit values 2**128 / 10**18 which is 340 quintillion 282 quadrillion 366 trillion 920 billion 938 million 463 thousand 463 nEUR max
```
djudjuu: that sounds like it might juuust be enough :)

contracts/ETO/ETOCommitment.sol
```diff
@@ -122,35 +125,35 @@ contract ETOCommitment is
...
+    // amount of new shares generated never exceeds number of tokens (uint128)
+    uint128 private _newShares;
     // platform fee in eth
-    uint112 private _platformFeeEth;
+    uint128 private _platformFeeEth;
     // platform fee in eur
-    uint112 private _platformFeeEurUlps;
+    uint128 private _platformFeeEurUlps;
+    // how many equity tokens goes to platform portfolio as a fee
+    uint128 private _tokenParticipationFee;
```
djudjuu: to stay more consistent we could also call this ```_tokenParticipationFeeAmount```

contracts/ETO/ETOTerms.sol
```diff
@@ -100,11 +100,15 @@ contract ETOTerms is
...
     // number of tokens that can be sold in whitelist
     uint256 public MAX_AVAILABLE_TOKENS_IN_WHITELIST;
+    // equity token scale power (10^precision)
+    uint256 public EQUITY_TOKEN_POWER;
 
     // base token price in EUR-T, without any discount scheme
     uint256 private TOKEN_PRICE_EUR_ULPS;
     // equity tokens per share
     uint256 private EQUITY_TOKENS_PER_SHARE;
+    // full price fraction
```
djudjuu: either the comment or the variable could be renamed for better understanding. maybe to ```DISCOUNTED_TOKEN_PRICE_EURO_ULPS``` or the comment to 
```// public-discounted price to be paid for a full token during the eto

File: contracts/ETO/ETOCommitment.sol [hunk 11]
```diff
@@ -916,58 +927,57 @@ contract ETOCommitment is
...
+        uint128 newTotalFixedSlotsTokenAmount = _totalFixedSlotsTokenAmount + uint128(fixedSlotEquityTokenAmount);
         // add new investor
-        uint32 newTotalInvestors = _totalInvestors + (firstTimeInvestment ? 1 : 0);
-        // new total eur equivalent invested
-        uint256 newTotalEquivEurUlps = _totalEquivEurUlps + equivEurUlps;
-        assert(newTotalEquivEurUlps < 2**112);
-        // resue eth/neur contribution slots for native currency bookkeeping
-        uint112 newTotalEth = _additionalContributionEth;
-        uint112 newTotalEurUlps = _additionalContributionEurUlps;
+        uint128 newTotalInvestors = _totalInvestors + (firstTimeInvestment ? 1 : 0);
```
djudjuu: why do we need to the unit of this from uint32 to uint128? 

the # of investors should not be affected by divisibility of the tokens...

contracts/ETO/ETOCommitment.sol
```diff
@@ -916,58 +927,57 @@ contract ETOCommitment is
...
+        // new total eur equivalent invested, disregard overflow for the same reason as with tokens
+        uint128 newTotalEquivEurUlps = uint128(_totalEquivEurUlps + equivEurUlps);
```
djudjuu: uint128 -> uint256 intended?

>  new total eur equivalent invested, disregard overflow for the same reason as with tokens
let's talk live about this!

contracts/ETO/ETOCommitment.sol
```diff
@@ -916,58 +927,57 @@ contract ETOCommitment is
...
     )
         private
     {
-        // total number of tokens must fit into uint56, uints are coerced to biggest size
-        uint256 newTotalTokensInt = _totalTokensInt + equityTokenInt;
-        assert(newTotalTokensInt < 2**56);
-        // as fixedSlotEquityTokenInt256 always lte equityTokenInt no chances to overflow
-        uint56 newTotalFixedSlotsTokensInt = _totalFixedSlotsTokensInt + uint56(fixedSlotEquityTokenInt);
+        // total number of tokens must fit into uint128, uints are coerced to biggest size
+        // as ticket token storage is 2**96, we need 2**32 tickets of max size to overflow so disregard
```
djudjuu: Note: Until I understand all storage layouts and assumptions to 110% I am sceptical of using such assumptions and would prefer to keep the check

contracts/ETO/ETOCommitment.sol
```diff
@@ -916,58 +927,57 @@ contract ETOCommitment is
...
         bool firstTimeInvestment
     )
         private
     {
-        // total number of tokens must fit into uint56, uints are coerced to biggest size
-        uint256 newTotalTokensInt = _totalTokensInt + equityTokenInt;
-        assert(newTotalTokensInt < 2**56);
-        // as fixedSlotEquityTokenInt256 always lte equityTokenInt no chances to overflow
-        uint56 newTotalFixedSlotsTokensInt = _totalFixedSlotsTokensInt + uint56(fixedSlotEquityTokenInt);
+        // total number of tokens must fit into uint128, uints are coerced to biggest size
```
djudjuu: > uints are coerced to biggest size

this is no longer accurate as the equity-tokenAmount is uint256, but is forced into uint128

contracts/ETO/ETOCommitment.sol
```diff
@@ -916,58 +927,57 @@ contract ETOCommitment is
...
         uint256 equivEurUlps,
         uint256 amount,
         bool firstTimeInvestment
     )
         private
     {
-        // total number of tokens must fit into uint56, uints are coerced to biggest size
-        uint256 newTotalTokensInt = _totalTokensInt + equityTokenInt;
-        assert(newTotalTokensInt < 2**56);
-        // as fixedSlotEquityTokenInt256 always lte equityTokenInt no chances to overflow
```
djudjuu: I dont get this comment, nor the new one.
new code is: 
uint 128 = uint128  + uint128

why can this not overflow?

contracts/ETO/ETOCommitment.sol
```diff
@@ -916,58 +927,57 @@ contract ETOCommitment is
...
+        _totalFixedSlotsTokenAmount = newTotalFixedSlotsTokenAmount;
         _totalInvestors = newTotalInvestors;
-        _totalEquivEurUlps = uint112(newTotalEquivEurUlps);
+        _totalEquivEurUlps = newTotalEquivEurUlps;
         _additionalContributionEth = newTotalEth;
         _additionalContributionEurUlps = newTotalEurUlps;
     }
 
-    function isCapExceeded(bool applyDiscounts, uint256 equityTokenInt, uint256 fixedSlotsEquityTokenInt, uint256 equivEurUlps)
+    function isCapExceeded(bool applyDiscounts, uint256 equityTokenAmount, uint256 fixedSlotsEquityTokenAmount, uint256 equivEurUlps)
```
djudjuu: this could with some documentation maybe:

> returns whether or not the cap would be exceeded if another investment defined by the given parameters would be made:
@param equityTokenAmount total amount of new equityTokens to be added
@param fixedSlotsEquityTokenAmount tokens of the above would be associated to reserved/fixed/pre-allocated slots
@param equivEurUlps the amount of EURO_TOKEN that is equiv to the newly-created tokens

contracts/ETO/ETOCommitment.sol
```diff
@@ -499,21 +503,20 @@ contract ETOCommitment is
             bool isEligible,
             uint256 minTicketEurUlps,
             uint256 maxTicketEurUlps,
-            uint256 equityTokenInt,
+            uint256 equityTokenAmount,
             uint256 neuRewardUlps,
-            bool maxCapExceeded,
-            uint256 spentAmountEurUlps
```
djudjuu: this should probably not have been removed

contracts/ETO/ETOCommitment.sol
```diff
@@ -690,44 +696,46 @@ contract ETOCommitment is
         // so if we sold MAX_AVAILABLE_TOKENS that were computed via calculateAmountWithoutFee from MAXIMUM_NUMBER_OF_TOKENS
         // the reverse operation, that is MAX_AVAILABLE_TOKENS + calculatePlatformTokenFee will not always be MAXIMUM_NUMBER_OF_TOKENS
         // while it's probably possible to detect it we put this check here
-        uint256 tokenParticipationFeeInt;
-        if (_totalTokensInt == MAX_AVAILABLE_TOKENS) {
-            uint256 maxTokens = tokenTerms.MAX_NUMBER_OF_TOKENS();
+        uint256 tokenParticipationFee;
+        uint256 maxTokens = tokenTerms.MAX_NUMBER_OF_TOKENS();
```
djudjuu: TODO from here

contracts/ETO/ETOTokenTerms.sol
```diff
@@ -85,14 +88,19 @@ contract ETOTokenTerms is Math, IContractId {
         EQUITY_TOKEN_NAME = equityTokenName;
         EQUITY_TOKEN_SYMBOL = equityTokenSymbol;
         EQUITY_TOKENS_PER_SHARE = equityTokensPerShare;
+        EQUITY_TOKENS_PRECISION = equityTokensPrecision;
+        EQUITY_TOKENS_POWER = 10**uint256(EQUITY_TOKENS_PRECISION);
+
```
djudjuu: > require(equityTokensPerShare > 0);

to avoid weird situations where 1 token is more than 1 share we could 

 ```require(equityTokensPerShare > EQUITY_TOKENS_POWER);

contracts/ETO/ETOTokenTerms.sol
djudjuu: > require(equityTokensPerShare > 0);

to avoid weird situations where 1 token is more than 1 share we could 

 ```require(equityTokensPerShare > EQUITY_TOKENS_POWER);```

djudjuu: in the beginning I was a bit confused about with the precision......so here we do mean the smallest unit of the equity token per share here, right?  i'll refer to them as ```whole equity ``` and ```mini-equity``` from now on. OR WE ME

1 Share = EQUITY_TOKENS_PER_SHARE * mini-Equity

(now, a bit later, i do think it makes NEVER sense to talk about whole-equity-tokens!)

contracts/ETO/ETOTokenTerms.sol
rudolfix: storage comes in 256bits chunks so increasing to 64bits allocates another WORD anyway. so we can go with 128bits and still use same amount of gas.

what do you mean by the same limit? if the token is 18 decimals then you need log2(10**18) ~59 bits to store one token and that is the increase

djudjuu: you're correct. I dont rmember how i got to that result. we need to have 2**56 * 2**59, so 128 is the next meaningful size, i agree

contracts/ETO/ETOCommitment.sol
rudolfix: it was totally caught by the test :) example: `ETOCommitment.js` `it("should calculate contribution", async () => {`

I removed it because for decimals 18 tokens this value is not relevant - will be 1 Ulps max. When calculating token amount we round down - for indivisible tokens you could buy whole tokens only. And difference between amount sent to the contract and amount spent could be quite big (up to token price!)

there's plenty comments on that in `ETOTerms::calculateContribution`

contracts/ETO/ETOCommitment.sol
rudolfix: let's go through the code of `updateTotalInvestment` again. there are indeed some "practical" assumptions like there will be less than 2^32 investments in single eto. we can add checked math/comments if things are not clear

we can add checked math everywhere, I prefer to track all arithmetic to be sure if it can overflow or not. asserting overflows can put you in dead end anyway...

contracts/ETO/ETOCommitment.sol
rudolfix: thx for the docstring

contracts/ETO/ETOTerms.sol
rudolfix: in other words we need to scale token amount to the TOKEN_POWER. and we multiply before we divide by price to minimize information lost by rounding down

contracts/ETO/ETOCommitment.sol
```diff
@@ -835,17 +843,20 @@ contract ETOCommitment is
...
+            fixedSlotEquityTokenAmount
         ) = ETO_TERMS.calculateContribution(investor, _totalEquivEurUlps, ticket.equivEurUlps, equivEurUlps, applyDiscounts);
         // kick out on KYC
         require(isEligible, "NF_ETO_INV_NOT_ELIGIBLE");
         // kick on minimum ticket and you must buy at least one token!
-        require(equivEurUlps + ticket.equivEurUlps >= minTicketEurUlps && equityTokenInt > 0, "NF_ETO_MIN_TICKET");
+        require(
+            equivEurUlps + ticket.equivEurUlps >= minTicketEurUlps &&
+            equityTokenAmount >= EQUITY_TOKEN_POWER &&
+            equityTokenAmount > 0, "NF_ETO_MIN_TICKET");
```
djudjuu: EQUITY_TOKEN_POWER will always be at least 1, so the second check is not needed

contracts/ETO/ETOCommitment.sol
djudjuu: this line has moved from inside to outside the if() condition, not necessary i think

contracts/ETO/ETOTokenTerms.sol
rudolfix: I think you are right, but to avoid any doubts we have two scales here
1. scale of the equity token
2. scale of the share vs. equity token

shares are always integers
equity tokens have from 0 to 18 decimals (which is technically called SCALE and I incorrectly call it PRECISION which I will correct now - it's making too much mess)

1 Share == EQUITY_TOKENS_PER_SHARE 
EQUITY_TOKENS_PER_SHARE  = 10**n * 10**decimals

where n is set up by issuer to have nice price of a token ~1EUR. in case of Fifth Force we had `n` == 4, greyp has n==6

btw. we do not have check for n to be integer in EQUITY_TOKENS_PER_SHARE and I will add this now

contracts/ETO/ETOTokenTerms.sol
rudolfix: yes this is a bug! the intention here was to require TPS > 0 as module division does not catch case of TPS == 0.

there's one more thing we should check: we require in the tests and we assume in many other places that TPS is power of 10. here we can have TPS == 2 tokens and all is fine (contract will work). any clever ideas to check if TPS is a power of ten without logarithm? we could use some loop certainly

contracts/ETO/ETOCommitment.sol
rudolfix: true, this is what you've get when you fix test cases without looking at the whole condition :)

contracts/ETO/ETOCommitment.sol
rudolfix: true, I had one invariant tested below which I dropped, now indeed it can go back

contracts/ETO/ETOTokenTerms.sol
```diff
@@ -88,10 +88,10 @@ contract ETOTokenTerms is Math, IContractId {
         EQUITY_TOKEN_NAME = equityTokenName;
         EQUITY_TOKEN_SYMBOL = equityTokenSymbol;
         EQUITY_TOKENS_PER_SHARE = equityTokensPerShare;
-        EQUITY_TOKENS_PRECISION = equityTokensPrecision;
-        EQUITY_TOKENS_POWER = 10**uint256(EQUITY_TOKENS_PRECISION);
+        EQUITY_TOKEN_DECIMALS = equityTokenDecimals;
+        EQUITY_TOKENS_POWER = 10**uint256(EQUITY_TOKEN_DECIMALS);
 
-        require(equityTokensPerShare >= 0 && equityTokensPerShare % EQUITY_TOKENS_POWER == 0, "NF_SHARES_NOT_WHOLE_TOKENS");
+        require(equityTokensPerShare > 0 && equityTokensPerShare % EQUITY_TOKENS_POWER == 0, "NF_SHARES_NOT_WHOLE_TOKENS");
```
djudjuu: > require(equityTokensPerShare > 0);

 we already have this chekc in line 76 so we can remove it from here entirely

contracts/ETO/ETOTokenTerms.sol
djudjuu: I think we could just set the way you described it in the other comment:
> EQUITY_TOKENS_PER_SHARE = 10n * 10decimals

so instead of 
```
EQUITY_TOKENS_PER_SHARE = equityTokensPerShare;
```
it would be
```
EQUITY_TOKENS_PER_SHARE = 10*full_tokens_per_share*10**DECIMALS;
```

contracts/ETO/ETOTerms.sol
djudjuu: why not have documentation? Alternatively, we could name the return parameter:

```
    function calculatePriceFraction(uint256 priceFrac) public constant returns(uint256 equivEuroUlps) {
        if (priceFrac == 1) {
            equivEuroUlps = TOKEN_PRICE_EUR_ULPS;
        } else {
            equivEuroUlps = decimalFraction(priceFrac, TOKEN_PRICE_EUR_ULPS);
        }
    }
```

contracts/ETO/ETOTokenTerms.sol
djudjuu: > I will add this now

I dont think you have added it yet. See this commment from me related to it: https://github.com/Neufund/platform-contracts/pull/330#discussion_r390266236

contracts/test/ETO/MockUncheckedETOTokenTerms.sol
```diff
@@ -20,7 +20,8 @@ contract MockUncheckedETOTokenTerms is ETOTokenTerms {
         uint256 maxNumberOfTokensInWhitelist,
         uint256 shareNominalValueUlps,
         uint256 shareNominalValueEurUlps,
-        uint256 equityTokensPerShare
+        uint256 equityTokensPerShare,
+        uint8 equityTokensPrecision
```
djudjuu: now that we renamed it to Decimals, we also need to rename it here

contracts/test/ETO/MockUncheckedETOTokenTerms.sol
```diff
@@ -46,5 +48,6 @@ contract MockUncheckedETOTokenTerms is ETOTokenTerms {
         TOKEN_PRICE_EUR_ULPS = tokenPriceEurUlps;
         MAX_NUMBER_OF_TOKENS_IN_WHITELIST = maxNumberOfTokensInWhitelist;
         EQUITY_TOKENS_PER_SHARE = equityTokensPerShare;
+
```
djudjuu: this extra newline sneaked in...

contracts/test/ETO/MockETOCommitment.sol
```diff
@@ -72,4 +72,14 @@ contract MockETOCommitment is
         emit LogTermsSet(msg.sender, address(etoTerms), address(equityToken));
         emit LogETOStartDateSet(msg.sender, startAt, logStartDate);
     }
+
+    //
+    // Override IAgreement internal interface to allow mocking up agreements for fixtures
+    //
+    function mCanAmend(address /*legalRepresentative*/)
```
djudjuu: why is this in here now? It is not related to the decimals issue, is is?

test/helpers/deployTerms.js
```diff
@@ -49,7 +52,7 @@ export const defEtoTerms = {
   ENABLE_TRANSFERS_ON_SUCCESS: false,
   INVESTOR_OFFERING_DOCUMENT_URL: "893289290300923809jdkljoi3",
   SHAREHOLDER_RIGHTS: null,
-  WHITELIST_DISCOUNT_FRAC: Q18.mul(0.3),
+  WHITELIST_DISCOUNT_FRAC: Q18.mul("0.3"),
```
djudjuu: do we need the quotes for it to work? If so, than there a bunch of other decimal numbers in line 15-22 for ```defaultShareholderTerms``` which need to be quoted too!

test/ETO/ETOCommitment.js
```diff
@@ -416,18 +428,36 @@ contract("ETOCommitment", ([, admin, company, nominee, ...investors]) => {
 
     it("with changing ether price during ETO", async () => {
       await skipTimeTo(publicStartDate.add(1));
+      // set some round ETH price to minimize rounding errors in the contract
+      await gasExchange.setExchangeRate(etherToken.address, euroToken.address, Q18, {
+        from: admin,
+      });
       const tokenprice = tokenTermsDict.TOKEN_PRICE_EUR_ULPS;
-      const ticket = Q18.mul(107.61721).add(1);
+      const ticket = Q18.mul("1070.61721").add(1);
```
djudjuu: There is an extra zero here (which doesnt really matter), but again, there are a lot of other places below where we multiply with a decimal-number and it is not using quotes

e.g. l. 1648, 1848, 2810 and prob more

test/ETO/ETOCommitment.js
```diff
@@ -416,18 +428,36 @@ contract("ETOCommitment", ([, admin, company, nominee, ...investors]) => {
...
-      expect(i1ticket[2].mul(2)).to.be.bignumber.eq(i2ticket[2]);
+      // but it does not mean that previous ticket amount is doubled due to rounding once, not twice
+      // i1ticket[2].mul(2) != i2ticket[2]
+      expect(
+        i1ticket[2]
+          .mul(2)
+          .sub(i2ticket[2])
+          .abs(),
+      ).to.be.bignumber.lt(2);
+      // this works because we set 1 eth == 2 eur
```
djudjuu: isnt' it 1 euro = 2 eth?
> .setExchangeRate(etherToken.address, euroToken.address ,2...

as a whole this test is not really easy to understand.

>       // token price increased twice
we mean it increased by factor of two, correct?￼

[2] refers to equityTokenAmount, so maybe a helpful comment could be: 
```// verify that same amount at double price buys almost same amount of equityToken (slightly less because of rounding)```

File: test/ETO/ETOCommitment.js [hunk 28]
```diff
@@ -416,18 +428,36 @@ contract("ETOCommitment", ([, admin, company, nominee, ...investors]) => {
...
+      const tokensI2 = ticket
+        .mul(2)
+        .mul(getTokenPower())
+        .divToInt(tokenprice);
+      expect(i2ticket[2]).to.be.bignumber.eq(tokensI2);
```
djudjuu: isnt' it 1 euro = 2 eth?
> .setExchangeRate(etherToken.address, euroToken.address ,2...




so in this test, we really verify that 

amountOfTokens * 2

test/ETO/ETOCommitment.js
```diff
@@ -416,18 +428,36 @@ contract("ETOCommitment", ([, admin, company, nominee, ...investors]) => {
...
         from: admin,
       });
       await investAmount(investors[2], ticket, "ETH", tokenprice);
       // token price increased twice
       const i1ticket = await etoCommitment.investorTicket(investors[1]);
       const i2ticket = await etoCommitment.investorTicket(investors[2]);
-      expect(i1ticket[2].mul(2)).to.be.bignumber.eq(i2ticket[2]);
+      // but it does not mean that previous ticket amount is doubled due to rounding once, not twice
+      // i1ticket[2].mul(2) != i2ticket[2]
+      expect(
```
djudjuu: as a whole this test is not really easy to understand.

>       // token price increased twice
we mean it increased by factor of two, correct?￼

[2] refers to equityTokenAmount, so maybe a helpful comment could be: 
```// verify that same amount at double price buys almost same amount of equityToken (slightly less because of rounding)```

File: test/ETO/ETOCommitment.js (hunk 28 above)
Comment (djudjuu) [inline test/ETO/ETOCommitment.js]: isnt' it 1 euro = 2 eth?
> .setExchangeRate(etherToken.address, euroToken.address ,2...




so in this test, we really verify that 

amountOfTokens * 2

Comment (djudjuu) [reply test/ETO/ETOCommitment.js]: @rudolfix I think this test is off. I think if we want to double the ehterprice the setExchangeRate-params need to be switched. 


Also I don't understand why there should be a ```.mul(2)``` here...

File: test/ETO/ETOCommitment.js [hunk 30]
```diff
@@ -1366,42 +1408,123 @@ contract("ETOCommitment", ([, admin, company, nominee, ...investors]) => {
...
+
+      return availableTokens;
+    }
+
+    function getOverflowPrice() {
+      // scale price so it will generate enough tokens to overlow 2**96
+      return Q18.div(Q18.div(defEquityTokenPower).mul(10000));
+    }
+
+    it("rejects on 2**96 tokens in ticket", async () => {
```
djudjuu: nice to have tests for this!

contracts/ETO/ETOTokenTerms.sol
djudjuu: Or we could do this:

```
    function is_power_of_ten(uint x) public pure returns(bool) {
        while (x > 9) {
            if (x % 10 == 0) { x = x / 10; }
            else { return false; }
        }
        return x == 1;
    }
}
```

test/ETO/ETOCommitment.js
```diff
@@ -1182,13 +1219,17 @@ contract("ETOCommitment", ([, admin, company, nominee, ...investors]) => {
   });
 
   describe("special ETO configurations", () => {
-    it("reverts overflow on internal equity token 56 bit", async () => {
-      // totals of equity tokens are stored in 56 bits number
-      const two = new web3.BigNumber(2);
-      const one = new web3.BigNumber(1);
+    it.skip("reverts overflow on internal equity token 128 bit", async () => {
+      // totals of equity tokens are stored in 128 bits number
+      // overflowing that is practically impossible, we'll need 2^32 tickets of 2^96 tickets (which is max)
```
djudjuu: > .... 2^32 tickets of 2^96 tickets ...

 should be of 2^96 tokens

test/ETO/ETOCommitment.js
```diff
@@ -2468,12 +2596,12 @@ contract("ETOCommitment", ([, admin, company, nominee, ...investors]) => {
...
         if (tokenTermsDict.MIN_NUMBER_OF_TOKENS.gt(totalInvestment[1])) {
           const missingTokens = tokenTermsDict.MIN_NUMBER_OF_TOKENS.sub(totalInvestment[1]);
-          let missingAmount = missingTokens.mul(tokenTermsDict.TOKEN_PRICE_EUR_ULPS);
+          let missingAmount = tokensCostEur(missingTokens, tokenTermsDict.TOKEN_PRICE_EUR_ULPS);
           if (missingAmount.lt(etoTermsDict.MIN_TICKET_EUR_ULPS)) {
             missingAmount = etoTermsDict.MIN_TICKET_EUR_ULPS;
           }
           // console.log(`min cap investment: ${missingTokens} ${missingAmount} EUR`);
-          await investAmount(investors[4], missingAmount, "EUR", tokenprice);
+          await investAmount(investors[4], missingAmount.add(1), "EUR", tokenprice);
```
djudjuu: why does it not work without the add(1)? is it because we do some rounding in ```tokenCostEur(...```

i can not let you get away with just adding 1 to make it work :) we need to at least have a theory why

test/ETO/ETOCommitment.js
```diff
@@ -4085,31 +4269,62 @@ contract("ETOCommitment", ([, admin, company, nominee, ...investors]) => {
...
+    const tranche2Tokens = tranche2Amount.mul(tokenPower).divToInt(tranche2Price);
+
+    // console.log(tranche1Tokens.add(tranche2Tokens).toString("10"));
+    const effectivePrice = totalAmount.mul(tokenPower).div(tranche1Tokens.add(tranche2Tokens));
+    // compare to our old linear approx of the same thing
+    expect(
+      calculateMixedTranchePriceLinear(totalAmount, tranche1Amount, tranche1Price, tranche2Price)
+        .sub(effectivePrice)
+        .abs(),
+    ).to.be.bignumber.lt(tokenPower.mul(2));
```
djudjuu: I dont understand this. Why is the linear approximation always less then two tokens off?

test/ETO/ETOTerms.js
```diff
@@ -63,9 +62,6 @@ contract("ETOTerms", ([, admin, investorDiscount, investorNoDiscount, ...investo
     expect((await durationTerms.contractId())[0]).to.eq(contractId("ETODurationTerms"));
     expect((await etoTerms.contractId())[0]).to.eq(contractId("ETOTerms"));
     expect((await etoTokenTerms.contractId())[0]).to.eq(contractId("ETOTokenTerms"));
-    for (const k of Object.keys(constTokenTerms)) {
-      expect(await etoTokenTerms[k]()).to.be.bignumber.eq(constTokenTerms[k]);
```
djudjuu: why are we not comparing those anymore?

test/ETO/ETOTerms.js
```diff
@@ -354,8 +388,9 @@ contract("ETOTerms", ([, admin, investorDiscount, investorNoDiscount, ...investo
       const maxTokenPrice = b2
         .pow("112")
         .div(tokenTerms.MAX_NUMBER_OF_TOKENS)
+        .mul(getTokenPower(tokenTerms))
         .floor();
-      // should pass
+      // should
```
djudjuu: accidental deletetion " pass"

test/ETO/ETOTerms.js
```diff
@@ -278,7 +308,7 @@ contract("ETOTerms", ([, admin, investorDiscount, investorNoDiscount, ...investo
       await expect(
         redeployTerms({
           MIN_TICKET_EUR_ULPS: modMinTicket,
-          PUBLIC_DISCOUNT_FRAC: Q18.sub(pubPriceFraction).add(1),
+          PUBLIC_DISCOUNT_FRAC: Q18.sub(pubPriceFraction.sub(10)),
```
djudjuu: why is it that before we increased the discount to make it fail, and now we decrease it to make it fail?

test/ETO/ETOTerms.js
djudjuu: because it no longer exists! :)

contracts/ETO/ETOTokenTerms.sol
rudolfix: I will go with `10*full_tokens_per_share*10**DECIMALS;`, really good idea!

contracts/test/ETO/MockETOCommitment.sol
rudolfix: yes, my fault but let me keep it - we'll use it on the backend fixtures generator to amend all agreements to autogenerated - important for testing

test/helpers/deployTerms.js
rudolfix: yes we should. but there's some trickery happening with number literals in javascript. there's no finite binary representation of 0.3 but still we obtain correct big number... I will fix wherever I can

btw. in Python there's no mercy, try this
```
>>> Decimal(0.3)
Decimal('0.299999999999999988897769753748434595763683319091796875')
```

test/ETO/ETOCommitment.js
rudolfix: you are right here! btw this test was insane to track all rounding errors. there's internal price conversion done with the inverse of the rate in the contract. setting price of ETH to 2 EUR removes rounding problems at that stage and I could write stricter test

test/ETO/ETOCommitment.js
rudolfix: I increased price of ETH by 100% so I can buy twice equity tokens for the same ETH amount. price of the token is in EUR.

and the last thing is indeed a trick. here's a comment I added
```
// this works because we set 1 eth == 2 eur so ticketEth.mul(2) is in fact currency conversion to EUR
      // on which amount of token is calculated
```

test/ETO/ETOCommitment.js
rudolfix: ok comment fixed

rudolfix: `// overflowing that is practically impossible, we'll need 2^32 investments with 2^96 ticket amount (which is max)`

test/ETO/ETOCommitment.js
rudolfix: good question! it's of course rounding but well understood. this is how price is computed that will let you cross min-cap
```
function tokensCostEur(tokens, price) {
    return tokens.mul(price).divToInt(getTokenPower());
  }
```
so we always round down here with `divToInt` which means that sometimes we'll land 1 wei from crossing the min cap and test fails.

test/ETO/ETOCommitment.js
rudolfix: this is also well understood. but it will take a diagram to explain. in ETOTerms contract we compute token amount and round for each tranche separately.

here we compute average weighted price of two tranches and ignore the gap. see `calculateContributionPrivate`

```
// calculate the really spent part to cover for rounding errors in next tranche (remainingAmount)
                    // commented out as with 18 digit scale we have 1 wei difference max and we'll not deploy any other
                    // precisions anymore
                    // discountedAmount = fixedSlotEquityTokenAmount * fixedSlotPrice / EQUITY_TOKEN_POWER;
```

why I even do it? to have very different way of calculating same value. not to just reproduce smart contract algo. to me this is a part of art of test writing.

PR #362, opened 2021-03-03
Proposal for simple exit contract

huesforalice:


contracts/Company/Extras/ExitController.sol
```diff
@@ -0,0 +1,277 @@
...
+        public
+        constant
+        returns (uint256)
+    {
+        if (_state == State.Setup ) {
+            return 0;
+        }
+        // calculate the amount of eligible proceeds based on the total equity token supply and the
+        // acquisition price
+        return Math.mul(_exitAquisitionPriceEurUlps, amountTokens) / _exitEquityTokenSupply;
```
banciur, line 139: Sadly I don't know about high precision math but there is also `Math.divRound` maybe it would be better her than simple `'`

contracts/Company/Extras/ExitController.sol
```diff
@@ -0,0 +1,277 @@
...
+        // we need a valid receiver address
+        require(newWallet != 0x0, "NF_INVALID_NEW_WALLET");
+        // we can only process wallets that have not been manually resolved yet
+        require(payoutManuallyResolved[lostWallet] == false, "NF_ALREADY_PAYED_OUT");
+        // only allow when timestamp has moved to next date
+        require(_manualPayoutResolutionStart < EQUITY_TOKEN.currentSnapshotId(), "NF_WAIT");
+
+        (uint256 _tokens, uint256 _proceeds) = eligibleProceedsForInvestor(lostWallet);
+        require(_proceeds > 0, "NF_NO_PROCEEDS");
+
```
banciur, line 225: Maybe it makes sense to check claims of `newWallet` and add requirement for KYC and bank account? It will help mitigate manual mistakes from nominee that will send those transactions.

banciur (commented):
More comments in conversation below PR

banciur:
Some loose ideas (more like brainstorming):
- I was thinking about manual phase. I think we should avoid manual contract operations as those are prone to copy paste errors  and write script that would take data as csv file and do it automatically.
- track addresses that would receive payout in manual process to not send eur twice to same address
  - this can be done offchain when preparing csv 
  - this might be valid usecase so maybe we should not prevent it
- Amount of EUR received from nominee will be known from start so it could be a constructor parameter and then validated when processing first transaction to make sure nominee would send correct amount
- Idea for frontend app. Highlight minimum redeem amount and tx price to let people know that with current gas price it doesn't make sense. This depends of structure of GPR owners.
- I wonder could we and if we can does it makes sense to incorporate redeem into manual payout. Then tx cost will be lower.
- Add minimum payout phase period. Just nice blockchainy phase to show we do it "right" and can't control everything.

contracts/Company/Extras/ExitController.sol
```diff
@@ -0,0 +1,277 @@
...
+        public
+        Agreement(universe.accessPolicy(), universe.forkArbiter())
+        Reclaimable()
+    {
+        UNIVERSE = universe;
+        EURO_TOKEN = UNIVERSE.euroToken();
+        EQUITY_TOKEN = equityToken;
+        _state = State.Setup;
+    }
+
```
arjunu, line 106:
```suggestion

////////////////////////
// External functions
////////////////////////
```

contracts/Company/Extras/ExitController.sol
huesforalice: The div round does a round up from .5 upwards, and I thought it would be safer to always round down the way / does. Or what do you think?

contracts/Company/Extras/ExitController.sol
huesforalice: I would say this is already done in the euro token controller contract in     function isTransferAllowedPrivate(address from, address to, bool allowPeerTransfers). So wallets that are not verified can't receive our-t. Could you double check if you think this is the case too?

desfero (approved):
Just small comment typos

contracts/Company/Extras/ExitController.sol
```diff
@@ -0,0 +1,281 @@
...
+        transitionTo(State.Payout);
+    }
+
+    // pay exit proceeds to an individual user
+    function payExitProceeds(address investor, uint256 equityTokenAmount)
+        private
+    {
+        // payout euro tokens to investor
+        uint256 _eligibleProceeds = eligibleProceedsForTokens(equityTokenAmount);
+        EURO_TOKEN.transfer(investor, _eligibleProceeds, "");
```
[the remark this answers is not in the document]
desfero: I think the same applies to manual payout too

desfero (changes requested):

contracts/Company/Extras/ExitController.sol
```diff
@@ -0,0 +1,281 @@
...
+
+/*
+To test:
+* Disbursal by nominee
+*
+ */
+
+contract ExitController is
+    KnownInterfaces,
+    Reclaimable,
```
desfero, line 21: Let's disallow reclaiming `EQUITY_TOKEN` completely and `EURO_TOKEN` up until we reach either `ManualPayoutResolution` or a new `Closed` state

contracts/Company/Extras/ExitController.sol
```diff
@@ -0,0 +1,281 @@
...
+
+        if (payoutManuallyResolved[investor]) {
+            return;
+        }
+
+        if (_state == State.Payout) {
+            equityTokens = EQUITY_TOKEN.balanceOf(investor);
+        }
+        else if (_state == State.ManualPayoutResolution) {
+            equityTokens = EQUITY_TOKEN.balanceOfAt(investor, _manualPayoutResolutionStart);
```
desfero, line 163: We can add support for users without claimed tokens here by interacting with the `ETOCommitment` smart-contract.
Pseudo-code 
```
  if (equityTokens == 0) {
                (claimedOrRefunded, equityTOkenAmount) = ETO_COMMITMENT.investorTicket(investor)

                if (claimedOrRefund == false) {
                    equityTokens = equityTOkenAmount;
                }
            }
           
```

